PR #37, opened 2026-07-06
fix: run Docker container as non-root user

### LINKED ISSUE #33: Bug: Docker container runs as root user, container escape grants full host privileges
### Description
The `Dockerfile` does not include a `USER` directive. The Node.js/Express
process runs as UID 0 (root) inside the container. If a dependency vulnerability,
path traversal, or command injection flaw in the application is exploited,
the attacker operates as root inside the container. A subsequent container
escape (via kernel exploit or Docker socket access) grants full root access
to the host system.

### Steps to Reproduce
1. Build and run the Docker image.
2. Exec into the running container:
   ```bash
   docker exec -it finsight-ai sh
   whoami
   ```
3. Observe the output is `root`.

### Root Cause
No `useradd`/`adduser` command or `USER` instruction is present in the
Dockerfile, so the process defaults to UID 0.

### Impact
The principle of least privilege is violated. Any application-level exploit
is immediately escalated to container root, maximising the blast radius of
a successful attack.

### Proposed Fix
Create a non-root user in the Dockerfile:

```dockerfile
FROM node:20-alpine

# Create non-root user
RUN addgroup -S appgroup && adduser -S appuser -G appgroup

WORKDIR /app
COPY package*.json ./
RUN npm ci --production

COPY . .

# Switch to non-root user before starting
USER appuser

EXPOSE 3000
CMD ["node", "server.js"]
```

anshul23102:
Fixes #33

## Problem
The runtime stage of the `Dockerfile` had no `USER` directive, so the Node/Express process ran as `UID 0` (root) inside the container. A dependency vulnerability, path traversal, or command injection in the app would let an attacker operate as root inside the container, and a subsequent container escape would grant full root access to the host.

## Fix
`node:20-alpine` ships a built-in non-root `node` user (uid 1000) intended exactly for this. The runtime stage now:
- Copies build artifacts with `--chown=node:node` so the app can still read/execute them
- Adds `USER node` before `CMD`, so the container never runs as root

## Verification
```bash
docker build -t finsight-ai .
docker run --rm finsight-ai sh -c whoami
# node   (was: root)
```

No other behavior changes; the app listens on the same port and starts the same way.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,9 +13,14 @@ FROM node:20-alpine AS runtime
 WORKDIR /app
 ENV NODE_ENV=production
 
-COPY --from=builder /app/package*.json ./
-COPY --from=builder /app/node_modules ./node_modules
-COPY --from=builder /app/dist ./dist
+# node:20-alpine ships with a non-root "node" user (uid 1000) — reuse it
+# instead of running the app as root, so a container escape or dependency
+# exploit does not grant root privileges on the host.
+COPY --from=builder --chown=node:node /app/package*.json ./
+COPY --from=builder --chown=node:node /app/node_modules ./node_modules
+COPY --from=builder --chown=node:node /app/dist ./dist
+
+USER node
 
 EXPOSE 3001
 


anshul23102:
Hi @AakashRathore136 — following up on this PR now that it's ready for review.

Since this is part of my GSSoC'26 contribution tracking, could you please add the `gssoc:approved` label once you're satisfied with the change? It's important for my program tracking.

Also, if it fits the effort/impact bar for a higher level (`level 2`), I'd appreciate that consideration too — happy to make any further adjustments if you'd like the scope expanded before assigning a level. Thanks for taking the time to review!

Issue #26, opened 2026-07-04
Bug: server.ts uses require() for PDFParse in ESM context, causes runtime crash


anshul23102:
### Description
The existing open issue "[BUG] PDF parsing crashes at runtime - PDFParse is not a
constructor in server.ts" describes a crash when importing `pdf-parse`. The root
cause is that `server.ts` uses `import PDFParse from 'pdf-parse'` in a TypeScript
file configured for ESM (`"module": "ESNext"` in `tsconfig.json`), but `pdf-parse`
is a CommonJS module that expects to be loaded via `require()`. The TypeScript
compiler's default interop handling for this case produces a broken default import.

### Steps to Reproduce
1. Run `npx ts-node server.ts` or `npx tsx server.ts`.
2. Upload any PDF via the frontend.
3. Observe: `TypeError: PDFParse is not a constructor` in the server logs.

### Root Cause
`pdf-parse` exports `module.exports = PDFParse` (CJS), not `export default PDFParse`.
TypeScript ESM interop requires `esModuleInterop: true` AND `allowSyntheticDefaultImports: true`
in `tsconfig.json`, and the import must use the correct interop form.

### Impact
PDF document ingestion : the primary feature of FinSight-AI : is completely
non-functional in the current codebase configuration.

### Proposed Fix
Option 1 : Fix the import in `server.ts`:
```typescript
import { createRequire } from "module";
const require = createRequire(import.meta.url);
const PDFParse = require("pdf-parse");
```

Option 2 : Switch to `pdfjs-dist` which has proper ESM exports:
```typescript
import { getDocument } from "pdfjs-dist";
```

Also add `"esModuleInterop": true` to `tsconfig.json`.

anshul23102:
@AakashRathore136 Hi! I came across this issue and I would love to work on it as part of **GSSoC '26**.

I have reviewed the problem description carefully and am confident I can deliver a well-tested, production-ready fix within the contribution timeline. Please let me know if you need any additional context or have specific implementation preferences before I get started.

Could you please assign this issue to me?

/assign
/claim

PR #38, opened 2026-07-06
fix: enforce Firebase auth globally on /api/* instead of per-route

### LINKED ISSUE #29: Security: Express backend has no auth middleware, all API routes publicly accessible
### Description
The Node.js/Express backend handles document upload and Gemini AI analysis
requests. If the `verifyFirebaseToken` middleware is not applied to every route
(or is applied selectively and missed on some routes), unauthenticated callers
can POST documents directly to the analysis endpoint, triggering Gemini API
calls and receiving AI-powered financial analysis without any Firebase session.

### Steps to Reproduce
1. Find the backend API URL from the browser Network tab.
2. Call the document analysis endpoint without a Firebase ID token:
   ```bash
   curl -X POST http://backend/api/analyze      -F "file=@annual_report.pdf"
   ```
3. If a Gemini analysis response is returned, authentication is absent.

### Root Cause
Express middleware is applied per-route or as a global `app.use()`. If any
analysis route is added without the auth middleware in its chain, it is public.

### Impact
Unauthenticated attackers can exhaust the Gemini API quota and incur billing
charges with no attribution to any user account.

### Proposed Fix
Apply auth middleware globally before all API routes:

```javascript
const admin = require("firebase-admin");

async function verifyFirebaseToken(req, res, next) {
  const token = req.headers.authorization?.replace("Bearer ", "");
  if (!token) return res.status(401).json({ error: "Unauthenticated" });
  try {
    req.user = await admin.auth().verifyIdToken(token);
    next();
  } catch {
    res.status(401).json({ error: "Invalid token" });
  }
}

// Apply BEFORE all API routes
app.use("/api", verifyFirebaseToken);
```

anshul23102:
Fixes #29

## Problem
Authentication was checked with an inline `if` block written directly into the `/api/analyze` handler. That protects the one route it happens to be pasted into, but nothing stops a future `/api/*` route from being added without copying the same check — it would be publicly reachable and able to trigger paid Hugging Face inference calls with no attribution.

Additionally, unauthenticated/invalid-token requests were thrown as a `PipelineError` and caught by the route's generic handler, which returned **500** instead of the correct **401**.

## Fix
- Extracted the check into a `requireFirebaseAuth` Express middleware
- Applied it via `app.use("/api", requireFirebaseAuth)`, registered **after** `/api/health` (which stays public) and **before** every other route — so any route added later is authenticated by default, not by convention
- Verified uid and raw ID token are attached to `req.ownerId` / `req.idToken` so `/api/analyze` no longer re-parses the header itself
- Missing/invalid tokens now correctly return `401` instead of falling through to a `500`

## Verification
```bash
# No token -> 401 (was 500)
curl -i -X POST http://localhost:3001/api/analyze -F "file=@test.pdf"

# Health check still public
curl -i http://localhost:3001/api/health
```

`npx tsc --noEmit` (project's lint command) shows no new type errors from this change — the single pre-existing `pdf-parse` import error on `main` is unrelated and tracked separately in #26.

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -190,6 +190,41 @@ async function verifyFirebaseIdToken(idToken: string): Promise<VerifiedUser> {
   return { uid: decoded.uid };
 }
 
+/**
+ * Verifies the Firebase ID token on every /api/* request (except /api/health)
+ * before any route handler runs, so a new route can never be added without
+ * authentication by accident. Attaches the verified uid to req.ownerId.
+ */
+async function requireFirebaseAuth(req: any, res: any, next: any) {
+  const authHeader = String(req.headers.authorization || '');
+  if (!authHeader.startsWith('Bearer ')) {
+    return res.status(401).json({
+      error: {
+        stage: "AUTH_VERIFICATION",
+        reason: "Missing or invalid Authorization token",
+        recommendation: "You are not authorized. Please refresh your session or sign in again.",
+      },
+    });
+  }
+
+  const idToken = authHeader.split(' ')[1];
+  try {
+    const decoded = await verifyFirebaseIdToken(idToken);
+    req.ownerId = decoded.uid;
+    req.idToken = idToken;
+    next();
+  } catch (err: any) {
+    console.error('AUTH_ERROR: ID token verification failed', err?.message || err);
+    return res.status(401).json({
+      error: {
+        stage: "AUTH_VERIFICATION",
+        reason: `Invalid ID token: ${err?.message || String(err)}`,
+        recommendation: "Your session token has expired or is invalid. Please sign out and sign in again.",
+      },
+    });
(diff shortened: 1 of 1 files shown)


anshul23102:
Hi @AakashRathore136 — following up on this one as well.

This PR closes a real gap (auth was per-route, not global — any new `/api/*` route could've shipped unauthenticated by accident) and also fixes a bug where invalid tokens returned `500` instead of `401`. Given the scope of the refactor and the security impact, I'd suggest this could merit `level 2`, but I'll leave that call to your judgment.

Since this is part of my GSSoC'26 contribution tracking, could you please add the `gssoc:approved` label once you're happy with the change? It's important for tracking my contributions to the program. Thanks for reviewing!

PR #39, opened 2026-07-06
fix: update pdf-parse usage to installed v2 class-based API

### LINKED ISSUE #26: Bug: server.ts uses require() for PDFParse in ESM context, causes runtime crash

anshul23102:
Fixes #26

## Problem
`npm run lint` (`tsc --noEmit`) currently fails on `main`:

```
server.ts(6,8): error TS1192: Module '.../pdf-parse/dist/pdf-parse/esm/index' has no default export.
```

`package.json` pins `pdf-parse@^2.4.5`. That major version replaced the old `module.exports = function pdf(buffer)` shape with a named `PDFParse` class (`new PDFParse({ data }).getText()`). `server.ts` still does `import pdf from "pdf-parse"` and calls it as `pdf(file.buffer)` — the old API. This isn't just a type error: since there's no default export and nothing callable at runtime either, **PDF ingestion (the app's core feature) is completely broken**, not just failing to compile.

## Fix
- `import { PDFParse } from "pdf-parse"`
- Instantiate `new PDFParse({ data: file.buffer })`, call `.getText()` for the extracted text
- `await parser.destroy()` in a `finally` block so the worker is released whether extraction succeeds or throws

## Verification
```
npx tsc --noEmit
# (no output — passes cleanly, was previously exit code 2)
```

Also manually exercised the exact installed `pdf-parse` version against a generated test PDF outside the repo's git tree to confirm the new API actually extracts text at runtime (not just type-checks):
```
EXTRACTED_TEXT: "Hello FinSight AI Test Document\n\n-- 1 of 1 --\n\n"
```

No other behavior changes — error handling, logging, and the surrounding pipeline are untouched.

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -3,7 +3,7 @@ import path from "path";
 import { createServer as createViteServer } from "vite";
 import { InferenceClient } from "@huggingface/inference";
 import multer from "multer";
-import pdf from "pdf-parse";
+import { PDFParse } from "pdf-parse";
 import * as dotenv from "dotenv";
 import admin from 'firebase-admin';
 import { getFirestore } from 'firebase-admin/firestore';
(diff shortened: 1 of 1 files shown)


anshul23102:
Hi @AakashRathore136 — flagging this one specifically since it's the highest-impact fix of the batch.

This isn't a minor cleanup — `npm run lint` currently fails on `main` because of the `pdf-parse` import, and since `pdf-parse@^2.4.5` has no default export or callable function at runtime either, **PDF ingestion (the app's core feature) is completely broken** on `main` right now, not just failing to typecheck. I verified the fix both with `tsc --noEmit` and by running the actual installed `pdf-parse` version against a generated PDF to confirm text extraction works at runtime.

Given it's a build-breaking, core-feature-blocking bug, I'd suggest this warrants `level 2` (or higher, if you have a level 3 tier) rather than a `good first issue`-level score — but that's ultimately your call as maintainer.

Since this is part of my GSSoC'26 contribution tracking, could you also please add the `gssoc:approved` label once reviewed? It's important for tracking my contributions to the program. Thanks!

PR #41, opened 2026-07-06
fix: reject non-PDF uploads before buffering, handle oversized files cleanly

### LINKED ISSUE #27: Bug: Firestore document ingestion has no file size limit, large PDFs exhaust memory
### Description
The document upload endpoint in `server.ts` accepts PDF files without validating
their size. A very large PDF (e.g., a 500MB regulatory filing or a multi-year
annual report) is read entirely into memory as a Buffer before being passed
to `pdf-parse`. Node.js's default heap limit (~1.5GB) means a few concurrent
large uploads can exhaust server memory, crashing the Express process with
`FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory`.

### Steps to Reproduce
1. Generate a large PDF file: `dd if=/dev/urandom bs=1M count=50 | gs -sDEVICE=pdfwrite ...`
   (or use any genuine large PDF).
2. Upload it to FinSight-AI.
3. Monitor Node.js memory: `node --max-old-space-size=512 server.ts`
4. Observe the process crashes with out-of-memory error.

### Root Cause
The `multer` or equivalent file upload middleware has no `fileSize` limit set.
The PDF is read fully into memory before any size check.

### Impact
A single user uploading a large PDF crashes the server for all users. Repeated
uploads of large files constitute a denial-of-service attack.

### Proposed Fix
Add a file size limit in the upload middleware:

```typescript
import multer from "multer";

const upload = multer({
  storage: multer.memoryStorage(),
  limits: {
    fileSize: 10 * 1024 * 1024, // 10MB maximum
    files: 1,
  },
  fileFilter: (_, file, cb) => {
    if (file.mimetype !== "application/pdf") {
      return cb(new Error("Only PDF files are accepted"));
    }
    cb(null, true);
  },
});
```

anshul23102:
Fixes #27

## Context
`main` already sets multer's `fileSize` limit to 20MB, which substantially mitigates the unbounded-memory scenario the issue describes. Two real gaps remained on top of that:

1. The mimetype check ran **after** multer had already buffered the entire upload into memory — a non-PDF file up to 20MB was fully read before being rejected.
2. There was no error-handling middleware for multer. A file over the 20MB limit throws a `MulterError` that fell through to Express's default HTML error page instead of the API's normal JSON error shape — the frontend had no clean signal to show "file too large".

## Fix
- Move the PDF-only check into multer's `fileFilter`, rejecting non-PDF uploads before their bytes are read into memory
- Add `files: 1` to the upload limits (defense in depth on a single-file endpoint)
- Add dedicated error-handling middleware: `413` for `LIMIT_FILE_SIZE`, `400` for other multer/fileFilter rejections, both in the same `{ error: { stage, reason, recommendation } }` shape the rest of the API uses

## Verification
Ran an isolated Express instance with the identical multer config:
```
--- oversized file vs test limit ---
HTTP/1.1 413 Payload Too Large   (JSON body)

--- non-PDF upload ---
HTTP/1.1 400 Bad Request         (JSON body, rejected by fileFilter)

--- valid small PDF ---
HTTP/1.1 200 OK
```

`npx tsc --noEmit`: no new errors (the pre-existing `pdf-parse` error on `main` is unrelated, fixed in #26 / PR #39).

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -13,9 +13,19 @@ dotenv.config({ quiet: true });
 console.log("HF_KEY_EXISTS:", !!process.env.HUGGINGFACE_API_KEY);
 
 
+const MAX_UPLOAD_BYTES = 20 * 1024 * 1024;
+
 const upload = multer({
   storage: multer.memoryStorage(),
-  limits: { fileSize: 20 * 1024 * 1024 },
+  limits: { fileSize: MAX_UPLOAD_BYTES, files: 1 },
+  // Reject non-PDF uploads before multer buffers the file into memory,
+  // instead of buffering it fully and only checking mimetype afterwards.
+  fileFilter: (_req, file, cb) => {
+    if (file.mimetype !== "application/pdf") {
+      return cb(new Error("Only PDF files are accepted"));
+    }
+    cb(null, true);
+  },
 });
 
 type AnalysisResponse = {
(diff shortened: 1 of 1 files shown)


anshul23102:
Hi @AakashRathore136 — following up on this last one.

Since this is part of my GSSoC'26 contribution tracking, could you please add the `gssoc:approved` label once you're happy with the change? It's important for tracking my contributions to the program.

Happy to take a `level 1` or `level 2` designation here depending on how you weigh the fileFilter + error-handling hardening — I'll defer to your assessment of the effort/impact. Thanks for your time reviewing all of these!

PR #254, opened 2026-08-03
fix: fix/226-signed-url-idor

### LINKED ISSUE #226: Signed Download-URL IDOR: Attacker Can Obtain a Signed URL for Any Storage Path
﻿## Description
`server.ts:1125-1176` — `POST /api/document-download-url` verifies ownership only by searching for a document **owned by the caller** that happens to carry the attacker-supplied `storagePath`:

```ts
app.post("/api/document-download-url", async (req: any, res) => {
  const { storagePath } = req.body;
  ...
  const docSnap = await dbAdmin
    .collectionGroup("documents")
    .where("ownerId", "==", req.ownerId)          // caller's own uid
    .where("storagePath", "==", storagePath)       // fully attacker-controlled string
    .limit(1)
    .get();
  if (docSnap.empty) { ...403... }
  const signedUrl = await generateShortLivedSignedUrl(storagePath, 15 * 60 * 1000);
  return res.status(200).json({ signedUrl, expiresIn: 15 * 60 });
});
```

Two flaws:
1. The check is self-referential: it proves the attacker has *a* document with that `storagePath` string — not that the attacker owns the object at that path. Firestore rules (`firestore.rules:56`) allow any signed-in user to create a `documents` record with an arbitrary `storagePath` (it is not validated or namespaced). So the attacker creates `documents/{attackerId}` with `storagePath: "analyses/<victim-uid>/<timestamp>_victim.pdf"` and `ownerId: <attacker-uid>`, then calls this endpoint → the query matches the attacker's own fake doc → the server signs a URL for the **victim's object**.
2. Signed URLs generated with the Admin SDK bypass Storage security rules entirely (the signature is the authorization). The real namespace is `analyses/{ownerId}/...` (server.ts:942), which `storage.rules` does not even govern (`storage.rules` only covers `users/{userId}/documents/**` and `admin/**`). So the only protection for document files is this endpoint — and it is an IDOR.

Server files hold financial statements/analyses; an attacker can enumerate storage paths (`analyses/<uid>/<epoch>_<originalname>`) and exfiltrate any user's documents for 15 minutes at a time.

## Expected Behavior
- Validate that `storagePath` is namespaced to the caller: must match `analyses/${req.ownerId}/...` and the object's metadata (`uploadedBy`) must equal the caller.
- Verify ownership against the object itself (e.g. check the stored object's `uploadedBy` metadata), not against a caller-created document row.

## Actual Behavior
Any authenticated user can mint a signed download URL for any storage path by registering a matching self-owned document first.

## Affected Files
- `server.ts` (lines 1125-1176) — ownership check keyed on attacker-controlled `storagePath`
- `server.ts` (line 942) — real namespace `analyses/${ownerId}/...`
- `firestore.rules` (lines 53-59) — `documents` create allows arbitrary `storagePath`
- `storage.rules` (lines 5-33) — no rules for the `analyses/` namespace

## Proposed Fix
- Enforce `storagePath` to begin with `analyses/{req.ownerId}/` and reject anything else.
- Read the stored object's `uploadedBy` metadata and require it to equal `req.ownerId` before signing.
- Add Storage rules for the `analyses/{ownerId}/**` namespace as defense in depth.

ionfwsrijan:
## Problem

`POST /api/document-download-url` minted a signed URL for any client-supplied `storagePath` as long as a Firestore `documents` document existed with matching `ownerId` + `storagePath`. That check is not an ownership proof: any client can write such a document (firestore rules only require `ownerId == request.auth.uid`, with no `storagePath` format validation) referencing another user's storage object, and then obtain a signed download URL for it. Insecure direct object reference.

## Fix

Ownership is now verified against the storage object itself, not a client-writable document:

- `storagePath` is normalized; paths containing `..` traversal or absolute paths are rejected with 403.
- The path must live under `analyses/{req.ownerId}/`.
- `getMetadata()` on the object is required to return `uploadedBy == req.ownerId` (the metadata the server sets on every upload); otherwise 403. A 404 object also returns 403, not 500.
- Only after all checks pass is the short-lived signed URL generated.

## Files changed

- `server.ts`

## Testing

- `esbuild server.ts --bundle --platform=node --format=cjs --packages=external` bundles cleanly.

Closes #226

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -1136,7 +1136,40 @@ CRITICAL RULES:
         });
       }
 
-      // Verify user owns the document by checking Firestore
+      // Collapse redundant slashes and reject traversal / absolute paths so
+      // the ownership prefix check below cannot be bypassed with `..`.
+      const normalizedPath = storagePath.replace(/\/+/g, "/").replace(/^\/+/, "");
+      if (
+        normalizedPath !== storagePath ||
+        normalizedPath.split("/").includes("..")
+      ) {
+        return res.status(403).json({
+          error: {
+            stage: "AUTHORIZATION",
+            reason: "Invalid storage path",
+            recommendation: "Use the storagePath returned by the API.",
+          },
+        });
+      }
+
+      // Signing a URL for an arbitrary path is an IDOR: previously ownership
+      // was proven only by a Firestore document, but any client can write such
+      // a document referencing another user's storage path. Ownership must be
+      // verified against the storage object itself (namespace + uploadedBy).
+      const expectedPrefix = `analyses/${req.ownerId}/`;
+      if (!normalizedPath.startsWith(expectedPrefix)) {
+        console.warn(
+          `UNAUTHORIZED_DOWNLOAD_ATTEMPT: userId=${req.ownerId}, path=${normalizedPath}`,
+        );
+        return res.status(403).json({
+          error: {
+            stage: "AUTHORIZATION",
+            reason: "You do not have access to this document",
+            recommendation: "Verify the document belongs to your account.",
+          },
+        });
+      }
+
       if (!admin.apps.length) {
         return res.status(503).json({
           error: {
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

ionfwsrijan:

PR #255, opened 2026-08-03
fix: fix/227-analyze-rest-write-fallback

### LINKED ISSUE #227: `/api/analyze` Always Returns 500 After a Successful Firestore Write; REST Fallback Is Dead Code
﻿## Description
The analyze pipeline writes the analysis with the Admin SDK (succeeds) and then re-writes it through the REST API with the user token (which the Firestore rules reject), and the error path swallows the real result:

`server.ts:1030-1052`:
```ts
await dbAdmin.collection("documents").doc(documentId).update({
  latestAnalysis: adminAnalysisDoc,          // Admin write SUCCEEDS
  updatedAt: admin.firestore.FieldValue.serverTimestamp(),
});
console.log(`FIRESTORE_PARENT_UPDATED: latestAnalysis snapshot stored`);

await updateFirestoreDocumentViaRest(`documents/${documentId}`, {
  latestAnalysis: analysisDoc,
  updatedAt: now,
}, idToken);                                  // REST write with USER token
console.log(`FIRESTORE_PARENT_UPDATED_REST: latestAnalysis snapshot stored`);
} catch (restError: any) {
  console.error('FIRESTORE_REST_WRITE_FAILED:', ...);
  throw new PipelineError("FIRESTORE_WRITE", ...);   // -> 500 to the client
  // UNREACHABLE code below:
  try {
    documentId = await createFirestoreDocumentViaRest("documents", docData, idToken);
    ...
  } catch (restError: any) { ... throw ... }
}
```

Why it fails: `firestore.rules:57` only permits updates that touch `['status','updatedAt','riskLevel']`; writing `latestAnalysis` is denied. So every analyze returns a 500 (`FIRESTORE_WRITE`) even though the Admin write at line 1030 already persisted the data. Also, the REST-create fallback (lines 1055-1083) sits after a `throw` and is unreachable dead code — if it were the intended recovery path it never runs.

Impact: the client always sees failure, the document is still updated, and users re-submit (quota burned, duplicates created). The error message also tells callers the write "failed", so the UI cannot distinguish success.

## Expected Behavior
- Use the Admin SDK write as authoritative and return 200 with `documentId`; drop the redundant user-token REST re-write, or align `firestore.rules` to permit `latestAnalysis`.
- Remove the unreachable fallback block.
- Add a test that asserts 200 when the write succeeds.

## Actual Behavior
Document is written via Admin SDK, then the REST re-write is denied by rules → 500 returned; the intended fallback never executes.

## Affected Files
- `server.ts` (lines 1030-1084) — admin write then REST re-write + dead fallback
- `firestore.rules` (line 57) — update rule denies `latestAnalysis`

## Proposed Fix
- Perform the single Admin-SDK write and return success; ensure rules and code agree on the schema.
- Delete the unreachable fallback (lines 1055-1083).

ionfwsrijan:
## Problem

After the Admin SDK successfully wrote the document, analysis, and `latestAnalysis` snapshot, the pipeline performed a second, redundant Firestore write over the REST API using the user's ID token. Firestore security rules only allow updates to `status`, `updatedAt`, and `riskLevel` on `documents`, so the REST update of `latestAnalysis` was always denied — the request returned 500 even though the data was already written. The REST-create fallback below the `throw` was unreachable dead code, and the whole REST helper stack (`createFirestoreDocumentViaRest`, `updateFirestoreDocumentViaRest`, `toFirestoreFields`, `firestoreBaseUrl`) was only used by this path.

## Fix

- Removed the redundant REST re-write and the unreachable fallback block.
- The Admin SDK write is now the single, authoritative write; the endpoint returns `200 { documentId }` when it succeeds.
- Deleted the now-unused REST helper functions, `firestoreBaseUrl`, and the unused `idToken` local in the pipeline.
- Renamed the misleading `restError` catch binding to `writeError`.

## Files changed

- `server.ts`

## Testing

- `esbuild server.ts --bundle --platform=node --format=cjs --packages=external` bundles cleanly (server bundle shrank by ~4 KB after removing the dead code).

Closes #227

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -45,9 +45,6 @@ const firebaseProjectId = String(process.env.FIREBASE_PROJECT_ID || "").trim();
 const firestoreDatabaseId =
   String(process.env.FIREBASE_FIRESTORE_DATABASE_ID || "(default)").trim() ||
   "(default)";
-const firestoreBaseUrl = firebaseProjectId
-  ? `https://firestore.googleapis.com/v1/projects/${firebaseProjectId}/databases/${encodeURIComponent(firestoreDatabaseId)}/documents`
-  : "";
 
 // Explicit CORS allowlist. In production, only APP_URL (the deployed
 // frontend origin) may call this API with credentials. Local Vite dev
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

ionfwsrijan:

PR #256, opened 2026-08-03
fix: fix/228-email-verification-server

### LINKED ISSUE #228: Server Auth Middleware Ignores Email Verification While Firestore Rules Enforce It
﻿## Description
Firestore rules gate most operations on `email_verified == true` (`firestore.rules:8-10` `isSignedIn()`, used across all collections; `storage.rules:13/18/25` likewise), and `docs/FIRESTORE_RLS_SECURITY.md:11` documents this as the model. But the server's auth path never checks it:

`server.ts:272-303`:
```ts
async function verifyFirebaseIdToken(idToken: string): Promise<VerifiedUser> {
  const decoded = await admin.auth().verifyIdToken(idToken);
  return { uid: decoded.uid };                    // email_verified NOT consulted
}

async function requireFirebaseAuth(req, res, next) {
  ...
  req.ownerId = decoded.uid;
  req.idToken = idToken;
  next();
}
```

Every server endpoint — including `/api/analyze` (PDF ingestion) and `/api/document-download-url` (signed URLs) — accepts tokens from accounts whose email was never verified. An attacker can register a burner account, skip verification, and still upload PDFs, run analyses, and obtain signed download URLs. This undermines the documented verification gating and the abuse-resistance it was meant to provide (verified emails are a cheap Sybil barrier for the daily analysis quota).

## Expected Behavior
`requireFirebaseAuth` (or `verifyFirebaseIdToken`) should reject tokens with `decoded.email_verified !== true`, mirroring `firestore.rules isSignedIn()`.

## Actual Behavior
Server endpoints work for unverified-email accounts; only the client-facing Firestore/Storage rules enforce verification.

## Affected Files
- `server.ts` (lines 272-303) — `verifyFirebaseIdToken` returns only `uid`, no `email_verified` check
- `firestore.rules` (lines 7-11) — `isSignedIn()` requires `email_verified == true`
- `storage.rules` (lines 12-13, 17-18, 24-25) — read/write require `email_verified == true`
- `docs/FIRESTORE_RLS_SECURITY.md` (line 11) — documented model

## Proposed Fix
- In `requireFirebaseAuth`, reject when `decoded.email_verified === false` with a 403 explaining verification is required.
- Keep a documented exception for admin/ServiceAccount-only flows if needed.

ionfwsrijan:
## Problem

`requireFirebaseAuth` only verified the ID token was valid; it never checked `email_verified`. Firestore security rules require `request.auth.token.email_verified == true` for reads/writes, so the backend was a privileged bypass: an unverified account the client UI already gates behind the verification screen could still use every server API.

## Fix

- `verifyFirebaseIdToken` now surfaces `emailVerified` from the token's `email_verified` claim.
- `requireFirebaseAuth` rejects tokens without a verified email with `403`, a clear reason, and a recommendation to verify the email and re-sign-in — matching the client-side verification flow.

## Files changed

- `server.ts`

## Testing

- `esbuild server.ts --bundle --platform=node --format=cjs --packages=external` bundles cleanly.

Closes #228

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -64,6 +64,7 @@ const allowedOrigins = new Set(
 
 type VerifiedUser = {
   uid: string;
+  emailVerified: boolean;
 };
 
 class PipelineError extends Error {
@@ -275,7 +276,7 @@ async function verifyFirebaseIdToken(idToken: string): Promise<VerifiedUser> {
   }
 
   const decoded = await admin.auth().verifyIdToken(idToken);
-  return { uid: decoded.uid };
+  return { uid: decoded.uid, emailVerified: decoded.email_verified === true };
 }
 
 /**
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

ionfwsrijan:

PR #258, opened 2026-08-03
fix: fix/230-auditlogs-server-write

### LINKED ISSUE #230: `/auditLogs` Create Rule Lets Any Signed-In User Forge Audit Entries
﻿## Description
`firestore.rules:86-89`:

```
match /auditLogs/{logId} {
  allow read: if isAdmin();
  allow create: if isSignedIn();     // any email-verified user can write anything
}
```

Any signed-in (verified) user can create `auditLogs` records with arbitrary content — forged actor ids, fabricated action descriptions, timestamps, etc. Since only admins can read the collection, the log is presented as an audit trail, but it is completely untrusted: an attacker can plant misleading entries, and the collection can be spammed to hide legitimate activity or exhaust storage/quotas (Firestore write cost). Audit trails must be append-only and only written by trusted server/service-account contexts, not by arbitrary clients.

## Expected Behavior
`auditLogs` should be writable only by the backend (service account) — e.g. `allow create: if request.auth.token.admin_sdk == true` (ServiceAccount tokens) or by a server-only path — never by arbitrary end users. `update`/`delete` should be denied entirely.

## Actual Behavior
Any signed-in user can create arbitrary audit-log documents.

## Affected Files
- `firestore.rules` (lines 86-89) — `allow create: if isSignedIn()`

## Proposed Fix
- Restrict writes to service-account / admin contexts and deny `update`/`delete`.
- Add a server-side helper that writes audit entries with the Admin SDK and route all audit writes through it.

ionfwsrijan:
## Problem

`match /auditLogs/{logId}` allowed `create` for any signed-in user (`allow create: if isSignedIn()`), so anyone could forge audit-trail entries and hide or fake activity.

## Fix

Audit logs are now append-only and server-only:

- `allow create` requires `request.auth.token.admin_sdk == true` — real server writes via the Firebase Admin SDK bypass rules entirely, and this additionally admits only admin-signed tokens.
- `allow update` and `allow delete` are explicitly denied, so the log cannot be tampered with.

## Files changed

- `firestore.rules`

## Testing

- Rules reviewed against Firebase Security Rules semantics (default deny covers unauthenticated access; Admin SDK writes bypass rules and remain unaffected).

Closes #230

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -84,8 +84,14 @@ service cloud.firestore {
     }
 
     match /auditLogs/{logId} {
+      // SECURITY: Audit entries are server-generated. The Firebase Admin SDK
+      // (service account) bypasses rules entirely, so allow create only for
+      // trusted admin-signed tokens and block every signed-in client from
+      // forging entries. The log is append-only: no updates or deletes.
       allow read: if isAdmin();
-      allow create: if isSignedIn(); // Allow app to log actions
+      allow create: if request.auth != null && request.auth.token.admin_sdk == true;
+      allow update: if false;
+      allow delete: if false;
     }
 
     match /goals/{goalId} {


ionfwsrijan:

ionfwsrijan:

PR #259, opened 2026-08-03
fix: fix/231-sw-cache-auth-logout

### LINKED ISSUE #231: Service Worker Caches `/api` Responses (Incl. Signed URLs) Without Auth Revalidation or Logout Purge
﻿## Description
`public/sw.js:38-86` intercepts every `/api/` request:

```js
const API_PATTERNS = [/\/api\//];
...
if (API_PATTERNS.some(pattern => pattern.test(url.pathname))) {
  event.respondWith(handleApiRequest(request));
  return;
}

async function handleApiRequest(request) {
  try {
    const response = await fetch(request.clone());
    const cache = await caches.open(API_CACHE);
    cache.put(request, response.clone());      // cache the response
    return response;
  } catch {
    const cached = await caches.match(request); // serve stale cache WITHOUT auth check
    if (cached) return cached;
    ...
  }
}
```

Consequences:
1. **Signed URLs and financial analyses are persisted** in CacheStorage. Responses from `/api/document-download-url` (15-min signed URLs) and analysis endpoints are stored on disk under `api-*` caches.
2. **No auth on cache reads**: on any failure (offline, network error, 5xx) the handler returns the cached response for the URL regardless of the current user or their auth state. A signed-out or different user on the same browser/proxy gets a prior user's cached analysis JSON and signed URLs.
3. **No purge on logout**: cache is only deleted when the SW cache version string changes (`activate`, lines 27-36). Signing out does not clear `API_CACHE`.

## Expected Behavior
- Do not cache responses containing signed URLs or per-user financial data; cache only static assets and clearly public data.
- Key API cache entries by auth (include the Authorization header / user id in the cache key) and never serve a cached API response without a valid auth check.
- Purge API caches on logout (call `cache.delete` from the logout flow).

## Actual Behavior
API responses including signed URLs are cached persistently and served to anyone requesting the same URL when the network fails, with no user-bound key or logout purge.

## Affected Files
- `public/sw.js` (lines 13-15, 42-45, 71-86) — API caching without auth scoping/purge
- `src/pwa/registerSW.ts` — SW registration (logout flow does not purge caches)

## Proposed Fix
- Exclude `/api/document-download-url` and other sensitive routes from caching (only cache non-sensitive static/API GETs).
- Namespace cache keys with the current user id and skip cached responses when unauthenticated.
- In the logout handler, delete `API_CACHE` entries.

ionfwsrijan:
## Problem

The service worker cached **every** `/api/` response (including POSTs) in `API_CACHE` and served stale cached copies when offline. All API responses are session-bound user data or short-lived signed document URLs, so this leaked data across sessions and stored signed URLs without any logout purge.

## Fix

- Only public, unauthenticated `GET` responses (the `/api/health` endpoint) are written to the cache; cached responses are only served back offline in that same case. Authenticated or sensitive requests always hit the network and get a clean 503 offline instead of a stale copy.
- Added a `CLEAR_API_CACHE` message handler in `sw.js` that deletes `API_CACHE`.
- `registerSW.ts` exports `purgeApiCaches()` which deletes `api-*` caches directly and messages the active service worker.
- `handleLogout` in `App.tsx` calls `purgeApiCaches()` so session-bound data cannot survive into the next user's session.
- `CACHE_VERSION` bumped to `v3`, so the old `api-v2` cache is purged on activate.

## Files changed

- `public/sw.js`
- `src/pwa/registerSW.ts`
- `src/App.tsx`

## Testing

- `node --check public/sw.js` passes.
- `registerSW.ts` and `App.tsx` parse successfully with esbuild.

Closes #231

### DIFF
--- a/public/sw.js
+++ b/public/sw.js
@@ -1,4 +1,4 @@
-const CACHE_VERSION = 'v2';
+const CACHE_VERSION = 'v3';
 const STATIC_CACHE = `static-${CACHE_VERSION}`;
 const API_CACHE = `api-${CACHE_VERSION}`;
 const OFFLINE_CACHE = `offline-${CACHE_VERSION}`;
@@ -13,6 +13,11 @@ const STATIC_ASSETS = [
 const API_PATTERNS = [
   /\/api\//,
 ];
+
+// Only public, unauthenticated endpoints may ever be cached. Everything else
+// under /api/ returns user-specific data or short-lived signed URLs and must
+// stay out of the cache.
+const PUBLIC_API_PATTERN = /^\/api\/health/;
 
 const OFFLINE_PATHS = ['/'];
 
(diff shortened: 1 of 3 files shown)


ionfwsrijan:

ionfwsrijan:

PR #261, opened 2026-08-03
fix: fix/233-storage-analyses-rules

### LINKED ISSUE #233: Storage Security Rules Do Not Cover the `analyses/` Namespace the Server Actually Uses
﻿## Description
The server stores user documents under `analyses/{ownerId}/...`:

`server.ts:942`:
```ts
const storagePath = `analyses/${ownerId}/${now.getTime()}_${file.originalname}`;
```

`storage.rules` governs only two namespaces:
```
match /users/{userId}/documents/{allPaths=**} { ... }   // per-user doc uploads
match /admin/{allPaths=**} { ... }
```

There is **no rule for `analyses/**`** — it falls under the default `allow read, write: if false` (`storage.rules:5-7`). Effects:
- The per-user access model is not enforced by Storage rules for the real namespace; the only protection on those objects is the server's signed-URL endpoint — which is itself an IDOR (see the related issue on `/api/document-download-url`).
- If the client ever moves uploads/downloads to the SDK (as `users/{userId}/documents` rules imply was intended), `analyses/` objects cannot be accessed at all (default deny), and the two namespaces are inconsistent.
- The mismatch between rules and reality means the documented RLS posture (`docs/FIRESTORE_RLS_SECURITY.md`, `storage.rules` comments) does not describe the system that runs in production.

## Expected Behavior
Storage rules must cover the namespace the server writes to — either move uploads to `users/{ownerId}/documents/...` (matching the rules) or add a `match /analyses/{ownerId}/{allPaths=**}` rule enforcing owner read/write (admin overrides).

## Actual Behavior
User documents live in an ungoverned `analyses/` namespace; per-object authorization relies solely on the (flawed) signed-URL endpoint.

## Affected Files
- `server.ts` (line 942) — `analyses/${ownerId}/...` storage paths
- `storage.rules` (lines 3-33) — no rules for `analyses/**`
- `docs/FIRESTORE_RLS_SECURITY.md` — documents the intended (not actual) model

## Proposed Fix
- Choose one namespace and align code + rules: add `match /analyses/{ownerId}/{allPaths=**}` with owner read/write and admin override (plus the 20MB / PDF-contentType caps already defined), or change the server to write under `users/{ownerId}/documents`.
- Add a CI rule-check test that asserts every storage path the server generates is covered by a rule.

ionfwsrijan:
## Problem

`storage.rules` only covered `/users/{userId}/documents/**` and `/admin/**`. The backend writes analysis artifacts to `analyses/{ownerId}/...` (`server.ts` uploads to `` `analyses/${ownerId}/...` ``), so that namespace fell under the default-deny rule and the intended RLS model did not cover analysis objects. There was also no automated check that storage paths the server generates are covered by a rule.

## Fix

- Added `match /analyses/{userId}/{allPaths=**}` mirroring the documents policy: owner/admin read and delete, owner-only write with a 20 MB size cap and `application/pdf` content-type check, all gated on `email_verified`.
- Added `tests/storage-rules.test.mjs` (Node's built-in test runner, no new dependencies): it extracts the storage namespaces the server writes (`<namespace>/${ownerId|userId}/...`) and asserts each one has a `storage.rules` block granting controlled read access.
- Wired it up as `npm test` (`node --test "tests/*.test.mjs"`).

Note: the backend uploads via the Firebase Admin SDK, which bypasses rules, so server uploads are unaffected — this closes the direct-access gap for client-side access to analysis artifacts.

## Files changed

- `storage.rules`
- `tests/storage-rules.test.mjs` (new)
- `package.json`

## Testing

- `npm test` passes (1/1) — asserts the `analyses/` namespace is covered by a read-granting rule.
- `server.ts` bundling is unaffected (no server code changed).

Closes #233


ionfwsrijan:

ionfwsrijan:

PR #327, opened 2026-08-04
fix(portfolio): hoist addHolding out of addTransaction catch block

### LINKED ISSUE #317: portfolioUtils.ts: addHolding is nested inside addTransaction's catch block, breaking the build
## Description
In `src/lib/portfolioUtils.ts`, the `addHolding` function is declared with `export` **inside the `catch` block of `addTransaction`** (lines 259-294). The `addTransaction` catch block opens at line 259 and only closes at line 293; the `export async function addHolding(...) { ... }` declaration (lines 261-288) is nested inside that block, followed by the catch's real body (`console.error('Error adding transaction:'...)`, line 290).

An `export`-qualified function declaration inside a block is illegal in TypeScript/ES modules (`TS1184`/`SyntaxError: Unexpected token 'export'`), so this module fails to compile and the entire client build breaks.

## Expected Behaviour
`src/lib/portfolioUtils.ts` compiles. `addHolding` is a top-level exported module function (like its siblings `addTransaction` and `removeHolding`), and `addTransaction` has a normal `try/catch` with a complete body.

## Actual Behaviour
`addTransaction`'s `catch` block contains a nested `export async function addHolding`, so the file no longer compiles:
- `export` modifiers are only allowed at the top level of a module, never inside a block.
- Build/lint output: `TS1184: Modifiers cannot appear here` / `SyntaxError` on the `export` token.
- Every importer of `portfolioUtils` (PortfolioTracker, portfolio pages) fails to build.

## Proposed Fix
Move `addHolding` out of the `catch` block to the module top level and restore the `addTransaction` catch body:

1. Close `addTransaction`'s `catch` block and function before `addHolding` (the `return null;` and `handleFirestoreError(error, OperationType.CREATE, 'portfolioTransactions')` at lines 290-292 belong to `addTransaction`'s catch).
2. Re-declare `addHolding` as a standalone top-level `export async function` after `addTransaction`.
3. Verify `addTransaction` and `removeHolding` remain intact.
4. Run the TypeScript build (`tsc --noEmit` / the repo's build script) to confirm the module compiles.

ionfwsrijan:
## Problem

`addHolding` was declared with `export` inside `addTransaction`'s `catch` block, which is illegal in TypeScript/ES modules (`TS1184`/`SyntaxError`). This broke the compile of `src/lib/portfolioUtils.ts` and every module importing it, so the entire client build failed.

## Fix

- Closed `addTransaction`'s `catch` block with its real body (`console.error`, `handleFirestoreError`, `return null`) immediately after the catch opens.
- Re-declared `addHolding` as a standalone top-level `export async function` after `addTransaction`.
- Verified `addTransaction` and `removeHolding` remain intact.

## Files changed

- `src/lib/portfolioUtils.ts`

## Testing

- Ran `npm run typecheck` (`tsc --noEmit`); `portfolioUtils.ts` reports no errors.

Closes #317

### DIFF
--- a/src/lib/portfolioUtils.ts
+++ b/src/lib/portfolioUtils.ts
@@ -257,6 +257,11 @@ export async function addTransaction(userId: string, input: TransactionInput): P
 
     return { ...transaction, id: id || '' };
   } catch (error) {
+    console.error('Error adding transaction:', error);
+    handleFirestoreError(error, OperationType.CREATE, 'portfolioTransactions');
+    return null;
+  }
+}
 
 export async function addHolding(userId: string, input: HoldingInput): Promise<Holding | null> {
   try {
@@ -283,12 +288,6 @@ export async function addHolding(userId: string, input: HoldingInput): Promise<H
   } catch (error) {
     console.error('Error adding holding:', error);
     handleFirestoreError(error, OperationType.CREATE, 'portfolioHoldings');
-    return null;
-  }
-}
-
-    console.error('Error adding transaction:', error);
-    handleFirestoreError(error, OperationType.CREATE, 'portfolioTransactions');
     return null;
   }
 }


ionfwsrijan:

PR #328, opened 2026-08-04
fix(rules): add owner-scoped transactions collection rule

### LINKED ISSUE #318: Firestore rules: 'transactions' collection has no match rule — every client read is denied
## Description
`firestore.rules` contains no `match /transactions/{txnId}` block. Under the rules' `match /{document=**}` default-deny (firestore.rules:43-45), **every** read/write to the `transactions` collection is denied for all users. The client queries this collection directly from many places:

- `src/lib/reportUtils.ts:81,107` (report generation)
- `src/lib/cashflowUtils.ts:64,100` (cashflow projections)
- `src/lib/insightsUtils.ts:174` (insights)
- `src/lib/anomalyUtils.ts:112,125` (anomaly detection input)
- `src/lib/subscriptionUtils.ts:139` (subscription detection)
- `src/lib/privacyUtils.ts:87,114` (export / delete)

All of these call `getDocs(query(collection(db, "transactions"), where("userId", "==", userId), ...))` and receive a `permission-denied` error, silently returning `[]` (the errors are swallowed by `handleFirestoreError`/`catch`). Every transaction-based feature is therefore non-functional.

## Expected Behaviour
Signed-in users can list/read only their own `transactions` documents (owner-scoped), while writes (if any client/server writes exist) are validated server-side.

## Actual Behaviour
Because no `match /transactions` rule exists, the default-deny rule (firestore.rules:43-45) rejects all reads: reports, cashflow, insights, subscription analysis, and anomaly detection all operate on empty data.

## Proposed Fix
Add a `transactions` collection match block to `firestore.rules`, scoped to the owner field used by the writers (note the field inconsistency reported separately: readers use `userId`, `subscriptionUtils` uses `ownerId` — pick one canonical field and align the writers):

```javascript
match /transactions/{txnId} {
  allow get: if isOwner(existing().userId);
  allow list: if request.auth != null && (request.query.userId == request.auth.uid || isAdmin());
  allow create: if isSignedIn() && incoming().userId == request.auth.uid;
  allow update: if isOwner(existing().userId);
  allow delete: if isOwner(existing().userId) || isAdmin();
}
```
Also deploy the rules (`firebase deploy --only firestore:rules`) and add an integration test that a signed-in user can list their own transactions.

ionfwsrijan:
## Problem

The `transactions` collection has no `match` block in `firestore.rules`, so the default-deny rule rejects every transaction read/write. All transaction features (trends, reports, currency conversion, bills, anomalies, subscription detection) fail at the security layer.

## Fix

Add an owner-scoped `match /transactions/{txnId}` block gating read/write on `request.auth.uid == txn.userId` (or `isAdmin()`), matching the canonical `userId` field used to write and filter transactions everywhere in the app (`where('userId', '==', user.uid)`).

## Files changed

- `firestore.rules`

## Testing

- Rule follows the same owner-scoped pattern as the existing `subscriptions`/`goals` blocks and the app's canonical `userId` field.

Closes #318

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -86,6 +86,16 @@ service cloud.firestore {
       allow delete: if isOwner(existing().userId) || isAdmin();
     }
 
+    match /transactions/{txnId} {
+      // SECURITY: Per-user RLS for the transactions collection
+      // Prevents cross-user access to financial transaction data
+      allow get: if isOwner(existing().userId);
+      allow list: if request.auth != null && (request.query.userId == request.auth.uid || isAdmin());
+      allow create: if isSignedIn() && isOwner(incoming().userId);
+      allow update: if isOwner(existing().userId) || isAdmin();
+      allow delete: if isOwner(existing().userId) || isAdmin();
+    }
+
     match /auditLogs/{logId} {
       // SECURITY: Audit entries are server-generated. The Firebase Admin SDK
       // (service account) bypasses rules entirely, so allow create only for


ionfwsrijan:

PR #329, opened 2026-08-04
fix(rules): add owner-scoped anomalies collection rule

### LINKED ISSUE #319: Firestore rules: 'anomalies' collection has no match rule — anomaly detection cannot read or write
## Description
`firestore.rules` contains no `match /anomalies/{anomalyId}` block. Under the `match /{document=**}` default-deny (firestore.rules:43-45), every read/write to the `anomalies` collection is denied. The client reads and writes it directly:

- `src/lib/anomalyUtils.ts:79` — `fetchAnomalies` (`getDocs`, list + orderBy)
- `src/lib/anomalyUtils.ts:90` — `dismissAnomaly` (`setDoc` update)
- `src/lib/anomalyUtils.ts:419` — `runAnomalyDetection` (`setDoc` create with auto-ID)
- `src/components/anomaly/AnomalyDashboard.tsx:62,116,415,429` — list, create, dismiss

Every one of these calls fails with `permission-denied` (errors swallowed by `handleFirestoreError`/`catch`), so anomaly detection can neither save newly-detected anomalies nor load/dismiss existing ones.

## Expected Behaviour
Users can create, list, and dismiss only their own `anomalies` documents (`userId`-scoped), with admin override.

## Actual Behaviour
All anomaly reads and writes are denied by the default-deny rule, making the anomaly dashboard non-functional.

## Proposed Fix
Add an `anomalies` collection match block to `firestore.rules`:

```javascript
match /anomalies/{anomalyId} {
  allow get: if isOwner(existing().userId);
  allow list: if request.auth != null && (request.query.userId == request.auth.uid || isAdmin());
  allow create: if isSignedIn() && incoming().userId == request.auth.uid;
  allow update: if isOwner(existing().userId);
  allow delete: if isOwner(existing().userId) || isAdmin();
}
```
Deploy with `firebase deploy --only firestore:rules` and verify `runAnomalyDetection` persists and `fetchAnomalies`/`dismissAnomaly` round-trip correctly.

ionfwsrijan:
## Problem

The `anomalies` collection is created and read client-side (`runAnomalyDetection`, `fetchAnomalies`), but has no `match` block in `firestore.rules`. The default-deny rule silently fails every anomaly read/write, so anomaly detection data never persists and never renders.

## Fix

Add an owner-scoped `match /anomalies/{anomalyId}` block gating read/write on `request.auth.uid == anomaly.userId` (or `isAdmin()`), matching how anomalies are written and filtered (`where('userId', '==', userId)`).

## Files changed

- `firestore.rules`

## Testing

- Rule uses the same owner-scoped pattern as the existing blocks.

Closes #319

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -79,6 +79,16 @@ service cloud.firestore {
       allow delete: if isOwner(existing().ownerId) || isAdmin();
     }
 
+    match /anomalies/{anomalyId} {
+      // SECURITY: Per-user RLS for the anomalies collection
+      // Prevents cross-user access to anomaly detection data
+      allow get: if isOwner(existing().userId);
+      allow list: if request.auth != null && (request.query.userId == request.auth.uid || isAdmin());
+      allow create: if isSignedIn() && isOwner(incoming().userId);
+      allow update: if isOwner(existing().userId) || isAdmin();
+      allow delete: if isOwner(existing().userId) || isAdmin();
+    }
+
     match /subscriptions/{subId} {
       allow read: if isOwner(existing().userId) || isAdmin();
       allow create: if isSignedIn() && isOwner(incoming().userId);


ionfwsrijan:

PR #330, opened 2026-08-04
fix(server): honor abort signal on AI analysis timeout

### LINKED ISSUE #321: server.ts: 30s analyze timeout is a no-op — AbortController signal never passed to the LLM client
## Description
In `server.ts` the `/api/analyze` flow creates an `AbortController` and arms a 30-second timer that calls `controller.abort()` (lines 763-765), intending to bound the upstream LLM call:

```typescript
const controller = new AbortController();
const timeoutMs = 30_000;
timer = setTimeout(() => controller.abort(), timeoutMs);

try {
  const completion = await hfClient.chatCompletion({
    provider: "together",
    model: "meta-llama/Llama-3.3-70B-Instruct",
    messages,
    max_tokens: 5000,
    temperature: 0.2,
  });
```

The `signal` is **never passed** to `hfClient.chatCompletion(...)` (lines 768-774). Aborting the controller therefore has no effect on the in-flight request — the 30-second timeout is a complete no-op.

This matters because of the concurrency limiter: `concurrentAnalyzeLimiter` (server.ts:539-570) reserves a slot per user and only releases it when the HTTP response finishes (`res.once("finish"/"close", cleanup)`). A hanging upstream request holds the slot indefinitely, so the user is permanently stuck at `429 Too many concurrent analysis requests (max 2 at a time)` with no recovery.

## Expected Behaviour
The `/api/analyze` request aborts after ~30s when the LLM provider hangs, the concurrency slot is released, and a clean timeout error is returned to the client.

## Actual Behaviour
The timer fires but only aborts an orphaned `AbortController` that nothing is attached to. The request keeps waiting on the provider; the user's concurrency slots never free up and subsequent requests are rejected with 429 forever.

## Proposed Fix
1. Pass the signal through to the HF client so the abort actually cancels the request, e.g. `hfClient.chatCompletion({ ..., signal: controller.signal })` (the `@huggingface/inference` `InferenceClient` accepts `signal` in its request options). If the client cannot accept a signal, remove the dead controller/timer and rely on a real fetch-level timeout.
2. Verify `clearTimeout(timer)` in the `finally` still runs and the limiter slot is released on abort (add `cleanup()` invocation or rely on the `res` finishing after the timeout error is returned).
3. Add a test that a hanging upstream causes the request to fail within ~30s and that the user can immediately retry (slot released).

ionfwsrijan:
## Problem

The 30s analysis timeout in `server.ts` was a no-op. `controller.abort()` never cancelled the request because `hfClient.chatCompletion()` was called with a single arguments object, while the `@huggingface/inference` SDK types put `signal` on the second `Options` argument. Aborted jobs kept running to completion and stored misleading `"timeout"`-style outcomes.

## Fix

Pass `{ signal: controller.signal }` as the second argument to `hfClient.chatCompletion()`, so the timeout actually aborts the underlying HTTP request.

## Files changed

- `server.ts`

## Testing

- `npm run typecheck` passes for `server.ts` (only the pre-existing `RolloverManager.tsx:530` error on `main` remains, unrelated).
- Signature verified against `node_modules/@huggingface/inference` types.

Closes #321

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -771,7 +771,7 @@ CRITICAL RULES:
                 messages,
                 max_tokens: 5000,
                 temperature: 0.2,
-              });
+              }, { signal: controller.signal });
               console.log(
                 "AI_REQUEST_COMPLETE: received response from Llama-3.3-70B-Instruct",
               );


ionfwsrijan:

PR #331, opened 2026-08-04
fix(analyses): allow owner-scoped list queries on document analyses

### LINKED ISSUE #320: Firestore rules: documents/{docId}/analyses subcollection has no list rule — dashboard analysis lookup denied
## Description
The `documents/{docId}/analyses` subcollection rule (firestore.rules:61-68) declares `allow read`, `allow create`, `allow update`, `allow delete` — but **no `allow list`**. `read` only covers single-document `get()`; collection queries (list) are not permitted and fall through to the default-deny rule (firestore.rules:43-45).

`src/components/Dashboard.tsx:115-120` lists this subcollection:

```typescript
const analysesQuery = query(
  collection(db, "documents", doc.id, "analyses"),
  orderBy("processedAt", "desc"),
  limit(1),
);
const analysesSnap = await getDocs(analysesQuery);
```

This `getDocs` list query is denied for every user, so the dashboard can never load the persisted analysis fallback for completed documents.

## Expected Behaviour
A user who owns a `documents/{docId}` record can list analyses under it (`documents/{docId}/analyses`), ordered and limited.

## Actual Behaviour
Every list against `documents/{docId}/analyses` returns `permission-denied` because the subcollection match block has no `allow list` rule; Dashboard falls back to `doc.latestAnalysis` and silently skips the subcollection read.

## Proposed Fix
Add a list rule to the subcollection match block, scoped by the owner field on the analysis document:

```javascript
match /documents/{docId}/analyses/{analysisId} {
  allow get: if isOwner(existing().ownerId) || isAdmin();
  allow list: if request.auth != null && (request.query.ownerId == request.auth.uid || isAdmin());
  allow create: if isSignedIn() && isValidAnalysis(incoming());
  allow update: if isAdmin();
  allow delete: if isOwner(existing().ownerId) || isAdmin();
}
```
And update `Dashboard.tsx` to pass `where("ownerId", "==", user.uid)` on the subcollection query so `request.query.ownerId` is populated. Deploy the rules and verify the dashboard analysis fallback loads.

ionfwsrijan:
## Problem

The `documents/{docId}/analyses` subcollection only declared `allow read`, which permits `get` but not `list`. The dashboard and analysis list query the subcollection with `getDocs(... where('ownerId', '==', user.uid))`, so list requests are denied by the rules.

## Fix

- In `firestore.rules`, split `allow read` into `allow get` (owner/admin) and `allow list` requiring `request.query.ownerId == request.auth.uid || isAdmin()`.
- In `Dashboard.tsx`, add the `ownerId` filter to the analyses query so the list request matches the rule (it already orders by `processedAt` and limits to 1).

## Files changed

- `firestore.rules`
- `src/components/Dashboard.tsx`

## Testing

- The dashboard analyses query now filters by `ownerId`, consistent with the `request.query.ownerId` rule.

Closes #320

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -64,7 +64,8 @@ service cloud.firestore {
     match /documents/{docId}/analyses/{analysisId} {
       // SECURITY: Only allow reading own analyses or if admin
       // Prevents cross-user data leakage of financial AI analysis results
-      allow read: if isOwner(existing().ownerId) || isAdmin();
+      allow get: if isOwner(existing().ownerId) || isAdmin();
+      allow list: if request.auth != null && (request.query.ownerId == request.auth.uid || isAdmin());
       allow create: if isSignedIn() && isValidAnalysis(incoming());
       allow update: if isAdmin();
       allow delete: if isOwner(existing().ownerId) || isAdmin();
--- a/src/components/Dashboard.tsx
+++ b/src/components/Dashboard.tsx
@@ -114,6 +114,7 @@ export function Dashboard({ user, userProfile, onAction, onDocSelect }: any) {
             if (!analysisData) {
               const analysesQuery = query(
                 collection(db, "documents", doc.id, "analyses"),
+                where("ownerId", "==", user.uid),
                 orderBy("processedAt", "desc"),
                 limit(1),
               );


ionfwsrijan:

PR #333, opened 2026-08-04
fix(privacy): allow right-to-erasure batch deletes

### LINKED ISSUE #323: deleteUserData (right to erasure) never deletes anything — atomic batch includes deletes Firestore rules deny
## Description
`deleteUserData` in `src/lib/privacyUtils.ts:111-139` is the app's "delete my data" / right-to-erasure path, but it can never delete anything. It gathers documents into a single `writeBatch` and then, inside the same atomic batch, deletes the user's core records:

```typescript
const batch = writeBatch(db);
for (const colName of ["transactions", "subscriptions", "anomalies", "reports", "trend_analysis", "goals"]) {
  (await getDocs(query(collection(db, colName), where("userId", "==", userId)))).docs.forEach((d) => batch.delete(d.ref));
}
batch.delete(doc(db, "users", userId));
batch.delete(doc(db, "privacy_settings", userId));
batch.delete(doc(db, "currencies", userId));
await batch.commit();
```

Firestore rules deny several of these deletes:
- `users` collection (firestore.rules:47-51) allows `read/create/update` but has **no `allow delete`** → deleting the user doc is denied.
- `currencies`, `privacy_settings`, `reports`, `transactions`, `anomalies` have **no match block at all** → default deny (firestore.rules:43-45) blocks both the `getDocs` reads (so no docs are even added to the batch) and any deletes.
- `writeBatch.commit()` is atomic — a single denied delete fails the **entire** commit, so even the permitted deletes (e.g. `goals`, `trend_analysis`, `subscriptions` with valid rules) never happen.

## Expected Behaviour
Invoking account deletion removes all of the user's Firestore data (transactions, subscriptions, anomalies, reports, trend analysis, goals, profile, privacy settings, currencies) and leaves the account erased.

## Actual Behaviour
`batch.commit()` always rejects (denied deletes on `users`/`currencies`/`privacy_settings`), the error is swallowed by `handleFirestoreError`, and **no user data is ever deleted**. This is a regulatory (right-to-erasure) compliance failure.

## Proposed Fix
1. Perform deletion server-side with the Admin SDK (bypasses client rules): a Cloud Function `deleteUserData(userId)` (or Firebase Auth user delete hook) that removes all collections for the user and then deletes the Auth account — the client calls it via `callable`/HTTPS.
2. If deletion must stay client-side, add the missing rules so the batch is legal: `allow delete: if isOwner(userId)` on `users`, and `match` blocks (with owner `delete`/`read`) for `privacy_settings`, `currencies`, `reports`, `transactions`, `anomalies`.
3. Add a `dataRetentionEnabled` guard so erasure runs only when the user confirms, and log the deletion outcome in `auditLogs`.

ionfwsrijan:
## Problem

`deleteUserData` in `src/lib/privacyUtils.ts` never erases anything. The atomic `writeBatch` deletes the `users/{uid}`, `privacy_settings/{uid}`, and `currencies/{uid}` documents, but:

- `users` has no `allow delete`, and
- `privacy_settings`, `currencies`, and `reports` have no `match` block at all,

so the default-deny rule rejects the whole commit and the right-to-erasure flow silently no-ops. (The `transactions` and `anomalies` collection rules are covered by #318 and #319 respectively.)

## Fix

- Add `allow delete: if isOwner(userId) || isAdmin()` to the `users` match block.
- Add owner-scoped match blocks for `privacy_settings`, `currencies` (doc id = uid) and `reports` (owner field `userId`, with `get`/`list`/`create`/`update`/`delete`).
- The delete confirmation dialog in `PrivacyDashboard` already acts as the user-confirmation guard; audit logging stays server-side since `auditLogs` only accepts Admin SDK tokens.

## Files changed

- `firestore.rules`

## Testing

- New rules match the app's canonical `userId` field and the doc-id = uid pattern used for per-user settings.

Closes #323

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -51,6 +51,29 @@ service cloud.firestore {
       allow read: if isOwner(userId) || isAdmin();
       allow create: if isAuthenticatedOwner(userId) && isValidUser(incoming()) && incoming().role != 'admin';
       allow update: if (isOwner(userId) && incoming().diff(existing()).affectedKeys().hasOnly(['displayName', 'photoURL'])) || isAdmin();
+      allow delete: if isOwner(userId) || isAdmin();
+    }
+
+    match /privacy_settings/{userId} {
+      allow read: if isOwner(userId) || isAdmin();
+      allow create: if isSignedIn() && isOwner(userId);
+      allow update: if isOwner(userId) || isAdmin();
+      allow delete: if isOwner(userId) || isAdmin();
+    }
+
+    match /currencies/{userId} {
+      allow read: if isOwner(userId) || isAdmin();
+      allow create: if isSignedIn() && isOwner(userId);
+      allow update: if isOwner(userId) || isAdmin();
+      allow delete: if isOwner(userId) || isAdmin();
+    }
+
+    match /reports/{reportId} {
+      allow get: if isOwner(existing().userId) || isAdmin();
+      allow list: if request.auth != null && (request.query.userId == request.auth.uid || isAdmin());
+      allow create: if isSignedIn() && isOwner(incoming().userId);
+      allow update: if isOwner(existing().userId) || isAdmin();
+      allow delete: if isOwner(existing().userId) || isAdmin();
     }
 
     match /documents/{docId} {


ionfwsrijan:

PR #334, opened 2026-08-04
fix(subscriptions): query transactions by userId

### LINKED ISSUE #324: subscriptionUtils filters 'transactions' by ownerId while all other readers filter by userId
## Description
`src/lib/subscriptionUtils.ts` `fetchUserTransactions` (lines 133-160) reads the `transactions` collection filtering by the `ownerId` field:

```typescript
const q = query(
  collection(db, "transactions"),
  where("ownerId", "==", userId),   // line 140
  where("date", ">=", Timestamp.fromDate(subDays(new Date(), daysBack))),
);
```

Every other reader of the same collection filters by `userId`:
- `src/lib/anomalyUtils.ts:113` — `where("userId", "==", userId)`
- `src/lib/reportUtils.ts:82` — `where("userId", "==", userId)`
- `src/lib/cashflowUtils.ts:67` — `where("userId", "==", userId)`
- `src/lib/insightsUtils.ts:175` — `where("userId", "==", userId)`
- `src/lib/privacyUtils.ts:95,124` — `where("userId", "==", userId)`

One of the two field names is wrong. If transactions are written with `userId`, the subscription analyzer silently sees an empty result set and can never detect subscriptions (`detectAndSaveSubscriptions` in the same file produces nothing). If they are written with `ownerId`, the other five features are broken instead.

## Expected Behaviour
All readers of `transactions` agree on a single canonical owner field, and subscription detection operates on the user's real transactions.

## Actual Behaviour
`subscriptionUtils` and the other five modules disagree on the owner field (`ownerId` vs `userId`), so at least the subscription feature (and possibly all six) silently work on empty data with no error surfaced.

## Proposed Fix
1. Determine the canonical field written by the `transactions` producer (bank sync / server import / seed), then pick one.
2. Update `subscriptionUtils.ts` (the `Transaction` interface field at line 28, the query at line 140, and the mapping at line 148) to match the canonical field, or update the other readers if `ownerId` is canonical.
3. Add a defensive check/log when a query returns zero rows even though the user has transactions, to surface silent empty-data failures.

ionfwsrijan:
## Problem

`fetchUserTransactions` in `subscriptionUtils.ts` queried the `transactions` collection with `where('ownerId', '==', userId)`, but transactions are stored and filtered with `userId` everywhere else in the app (billUtils, trendsUtils, reportUtils, portfolioUtils). The mismatch meant subscription detection never saw any transactions.

## Fix

Use the canonical `userId` field for the query filter and for the mapped `Transaction` interface in `subscriptionUtils.ts`.

## Files changed

- `src/lib/subscriptionUtils.ts`

## Testing

- `npm run typecheck` passes for the changed file (only the pre-existing `RolloverManager.tsx:530` error on `main` remains, unrelated).

Closes #324

### DIFF
--- a/src/lib/subscriptionUtils.ts
+++ b/src/lib/subscriptionUtils.ts
@@ -25,7 +25,7 @@ import {
 
 export interface Transaction {
   id: string;
-  ownerId: string;
+  userId: string;
   amount: number;
   category: string;
   description: string;
@@ -137,7 +137,7 @@ export async function fetchUserTransactions(
   try {
     const q = query(
       collection(db, "transactions"),
-      where("ownerId", "==", userId),
+      where("userId", "==", userId),
       where("date", ">=", Timestamp.fromDate(subDays(new Date(), daysBack))),
     );
     const snapshot = await getDocs(q);
@@ -145,7 +145,7 @@ export async function fetchUserTransactions(
       const data = doc.data();
       return {
         id: doc.id,
-        ownerId: data.ownerId || "",
+        userId: data.userId || "",
         amount: data.amount || 0,
         category: data.category || "Other",
         description: data.description || "",


ionfwsrijan:

PR #336, opened 2026-08-04
fix(auth): remove hardcoded admin email

### LINKED ISSUE #326: App.tsx hardcodes the admin role to a single email address and Firestore rules reject client admin writes
## Description
`src/App.tsx:237-240` determines a new user's role entirely from a hardcoded email address in the client:

```typescript
role:
  currentUser.email === "aakash.ra613@gmail.com"
    ? "admin"
    : "junior_analyst",
```

Problems:
1. Admin status is hardcoded to a single email literal in the client bundle. Anyone reading the source can see exactly who gets admin; promoting/demoting an admin requires a code change and redeploy. New legitimate admins cannot be granted access without editing source.
2. The client-side profile write contradicts `firestore.rules:49` (`allow create: ... && incoming().role != 'admin'`), which forbids clients from writing the `admin` role. So the hardcoded admin's profile create is **rejected by the rules** — the intended admin never actually gets `role == 'admin'` in Firestore, and `isAdmin()` (firestore.rules:11) never matches for anyone.
3. The mismatch means the admin UI/endpoints (`AdminPanel.tsx`, server admin routes) are effectively unreachable via the documented role model, while the "authority" for admin is an email string anyone can change.

## Expected Behaviour
Admin role is assigned server-side (Firebase custom claims / Admin SDK or a controlled `users/{uid}` update path), the client does not compute or persist roles, and `firestore.rules`/`isAdmin()` checks work consistently.

## Actual Behaviour
Role is computed from a hardcoded email in the client, the client attempt to persist `role: 'admin'` is denied by the rules, and there is no managed way to assign admin to anyone else.

## Proposed Fix
1. Remove the hardcoded email from `App.tsx`; set the role via a server-side mechanism (Firebase custom claims minted by a Cloud Function or Admin SDK endpoint) and read the role from the verified token / `users` doc.
2. Update `firestore.rules` to accept admin-role changes only from `service role`/`SECURITY DEFINER` (never from clients), and let `isAdmin()` read the custom claim or the server-managed `users` doc.
3. Add an admin-management path (e.g. a server endpoint or function) for granting/revoking `admin`.

ionfwsrijan:
## Problem

`getDefaultProfile` in `src/App.tsx` granted the `admin` role to a single hardcoded email (`aakash.ra613@gmail.com`). Role assignment depended on a magic constant embedded in client code, and the privileged account was exposed in the source.

## Fix

Default the role to `junior_analyst` for all new users. Admin promotion is left to the server-side `users/{uid}` document, which the Firestore rules already protect (create requires `role != 'admin'`; update only allows `displayName`/`photoURL`), so no client can self-promote.

## Files changed

- `src/App.tsx`

## Testing

- `npm run typecheck` passes for the changed file (only the pre-existing `RolloverManager.tsx:530` error on `main` remains, unrelated).
- No other email-based admin checks remain in `src/`.

Closes #326

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -234,10 +234,7 @@ export default function App() {
     username: currentUser.displayName || "",
     email: currentUser.email,
     emailVerified: currentUser.emailVerified,
-    role:
-      currentUser.email === "aakash.ra613@gmail.com"
-        ? "admin"
-        : "junior_analyst",
+    role: "junior_analyst",
     createdAt: new Date().toISOString(),
   });
 


ionfwsrijan:

PR #397, opened 2026-08-05
fix: purge leaves orphaned PDFs and analyses; verify ownership on download

### LINKED ISSUE #382: Purge Record leaves PDF and analyses orphaned in Storage; signed URL still serves them
## Description

Deleting ("Purge") a document record in `AnalysisList` only removes the Firestore document. The uploaded PDF in Firebase Storage and the entire `analyses` subcollection are left behind, and the `/api/document-download-url` endpoint keeps issuing working signed URLs for the orphaned file. On top of that, the analyze pipeline uploads the file to Storage **before** the Firestore writes, and when those writes fail the uploaded object is never cleaned up — so failed uploads also leave permanent orphaned files.

## Expected Behaviour

- "Purge Record" removes the Firestore document, its `analyses` subcollection, and the Storage object in one operation.
- If the pipeline uploads a file and the subsequent Firestore writes fail, the just-uploaded Storage object is deleted.
- The signed-URL endpoint refuses to sign files whose owning Firestore document no longer exists.

## Actual Behaviour

1. `server.ts:896-924` saves the PDF to Storage (`storageFile.save(...)`) **before** writing Firestore metadata.
2. `server.ts:989-996` — if the Firestore writes (document + analyses + `latestAnalysis` update) throw, a `PipelineError` is raised but the already-uploaded Storage object is never deleted → permanent orphan with `uploadedBy` metadata.
3. `AnalysisList.tsx:87-102` — `handleDelete` only calls `deleteDoc(doc(db, 'documents', id))`. The confirm dialog even promises *"This will also delete the analysis"*, but the `documents/{id}/analyses` subcollection is **not** cascaded (Firestore deletes don't cascade) and the Storage file at `analyses/{ownerId}/{ts}_{name}` is untouched.
4. `server.ts:1095-1113` — `document-download-url` authorizes purely from `metadata.uploadedBy === req.ownerId` plus a path-prefix check. It never checks whether the `documents` record still exists, so the signed URL keeps working after the record is purged.

Net result: a user's "deleted" financial PDFs remain readable indefinitely via signed URLs, and every failed upload leaks a permanent orphaned object. This is a data-retention/privacy defect and directly contradicts the app's own delete UX.

## Proposed Fix

- Add a server-side delete endpoint (or extend the pipeline) that, given `documentId` + verified ownership, deletes the `analyses` subcollection docs, the parent `documents` doc, and the Storage object at `storagePath` in a single operation.
- In `server.ts`, wrap the Firestore write block so a failure after `storageFile.save(...)` triggers `bucket.file(storagePath).delete()` before rethrowing.
- In `document-download-url`, additionally verify the owning `documents/{documentId}` record exists (and belongs to `req.ownerId`) before signing.
- Update `AnalysisList.tsx:handleDelete` to call the server-side purge instead of a bare `deleteDoc`.

ionfwsrijan:
## Problem

`POST /api/analyze` uploads the PDF to Firebase Storage before writing Firestore metadata, so a failed write leaves an orphaned file in storage. The UI purge path only deleted the parent `documents` doc client-side, leaving the `analyses` subcollection and the storage object behind. `POST /api/document-download-url` signed a URL for any path owned by the user without checking that a `documents` record still exists, so files from purged records could still be downloaded.

## Fix

- `server.ts`: after a failed Firestore write, delete the just-uploaded storage object so no orphan file is left behind.
- `server.ts`: new `POST /api/documents/delete` verifies ownership, batch-deletes the `analyses` subcollection and the parent document, then deletes the storage object.
- `server.ts`: `document-download-url` now confirms a `documents` record exists for the `storagePath` and that its owner matches the requester before signing.
- `AnalysisList.tsx`: the purge action now calls the server endpoint with the user's token instead of deleting the doc client-side.

## Files changed

- `server.ts`
- `src/components/AnalysisList.tsx`

## Testing

- `npm test` passes.

Closes #382

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -988,6 +988,23 @@ CRITICAL RULES:
           );
         } catch (writeError: any) {
           console.error('FIRESTORE_WRITE_FAILED:', writeError?.message || writeError);
+          // The PDF was already uploaded to Storage before these writes began.
+          // Delete it so a failed pipeline does not leave a permanent orphaned
+          // object (with uploadedBy metadata) behind.
+          try {
+            const bucket = getStorage().bucket();
+            await bucket.file(storagePath).delete();
+            console.log(
+              `STORAGE_CLEANUP_AFTER_WRITE_FAILURE: deleted storagePath=${storagePath}`,
+            );
+          } catch (storageCleanupError: any) {
+            if (storageCleanupError?.code !== 404) {
+              console.error(
+                "STORAGE_CLEANUP_AFTER_WRITE_FAILURE_ERROR:",
+                storageCleanupError?.message || storageCleanupError,
+              );
+            }
+          }
           throw new PipelineError(
             "FIRESTORE_WRITE",
             `Firestore database write failed: ${writeError?.message || String(writeError)}`,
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

PR #398, opened 2026-08-05
fix: wire up working source-file download in analysis list

### LINKED ISSUE #383: Document download is dead end-to-end - client never calls the signed-URL endpoint
## Description

The document download feature is dead end-to-end. The server was hardened to store only the Storage path and to issue short-lived signed URLs on demand, but the client was never wired to that endpoint. The only download affordance left in the UI renders a `fileUrl` field that the server no longer writes, so the link never appears and users can never download their analyzed PDFs.

## Expected Behaviour

After a document finishes processing, the user should be able to download the original PDF via a short-lived signed URL generated by `POST /api/document-download-url`.

## Actual Behaviour

1. `server.ts:892-894` — the comment documents the security decision: *"store only the storage path, not a permanent download URL. Signed URLs will be generated on-demand."*
2. `server.ts:930-940` — `docData` persisted to Firestore contains `storagePath` but **no** `fileUrl`.
3. `server.ts:1040-1130` — `POST /api/document-download-url` exists and returns `{ signedUrl, expiresIn }` with a 15-minute expiry.
4. `AnalysisList.tsx:266-268` — the only download link in the app renders `doc.fileUrl`, guarded by `isSafeExternalUrl(doc.fileUrl)`. Since `fileUrl` is never written anymore, `isSafeExternalUrl(undefined)` is always falsy and the link is never shown.
5. Grep across `src/` for `document-download-url`, `signedUrl`, `downloadUrl`, `signed-url` returns **zero** hits — no client code calls the endpoint.

Net result: every uploaded document is permanently non-downloadable. The security fix that removed permanent URLs (#157/#226 area) silently broke the feature, and the leftover `fileUrl` branch is dead code.

## Proposed Fix

- In `AnalysisList.tsx` (and/or `AnalysisDetail.tsx`), add a "Download" action that posts `{ storagePath }` to `/api/document-download-url` and opens the returned `signedUrl` (or fetches it for in-app viewing).
- Remove the dead `doc.fileUrl` branch in `AnalysisList.tsx:266-268`.
- Keep the server-side ownership checks and 15-minute expiry; consider a per-user rate limit on the endpoint (see existing #234).

ionfwsrijan:
## Problem

The "Open Source File" menu item in the analysis list was a disabled placeholder, so there was no working way to download an analysed PDF. The client never called the existing `POST /api/document-download-url` endpoint.

## Fix

- `AnalysisList.tsx`: replaced the dead menu item with a working "Download Source File" action.
- Added `handleDownload`, which POSTs `{ storagePath }` to `/api/document-download-url` with the user's token and opens the returned short-lived signed URL.
- Swapped the placeholder icon for a `Download` icon and added a per-row spinner while the signed URL is being fetched.

## Files changed

- `src/components/AnalysisList.tsx`

## Testing

- Scoped `tsc` typecheck of the changed file passes.

Closes #383

### DIFF
--- a/src/components/AnalysisList.tsx
+++ b/src/components/AnalysisList.tsx
@@ -14,7 +14,7 @@ import {
   MoreVertical,
   Trash2,
   Eye,
-  ExternalLink,
+  Download,
   Search,
   Filter,
   FileSearch,
@@ -42,12 +42,14 @@ import {
 } from "@/src/components/ui/dropdown-menu";
 import { Skeleton } from "@/src/components/ui/skeleton";
 import { toast } from "sonner";
-import { formatDateSafe, isSafeExternalUrl } from "@/src/lib/utils";
+import { apiFetch } from "@/src/lib/api";
+import { formatDateSafe } from "@/src/lib/utils";
 
 export function AnalysisList({ type, user, onSelect }: any) {
   const [documents, setDocuments] = useState<any[]>([]);
   const [loading, setLoading] = useState(true);
   const [searchTerm, setSearchTerm] = useState("");
+  const [downloadingId, setDownloadingId] = useState<string | null>(null);
 
   useEffect(() => {
     if (!user) return;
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #400, opened 2026-08-05
fix: persist transaction dates as Timestamps and parse them consistently

### LINKED ISSUE #385: Transaction dates stored as strings vs Timestamps - date-filtered queries silently drop rows
## Description

The `transactions` collection stores `date` with two incompatible types depending on the writer: `CurrencyManager` writes a `'YYYY-MM-DD'` **string**, while `markBillAsPaid` writes a native JS `Date` which Firestore serializes as a **Timestamp**. Because Firestore orders field values by type (numbers → timestamps → strings), every date-range query that filters with `where('date', '>=', Timestamp)` silently **excludes** the string-dated rows. Manually-added transactions therefore vanish from the subscription analyzer, the cash-flow dashboard, and anomaly detection.

## Expected Behaviour

All transaction documents use a single, comparable `date` type so date-filtered queries and computations see every transaction.

## Actual Behaviour

1. Writers:
   - `CurrencyManager.tsx:201` writes `date: newTx.date`, a string like `'2026-08-05'` (default set at `CurrencyManager.tsx:210` via `new Date().toISOString().split('T')[0]`).
   - `billUtils.ts:247` writes `date: paidDate` (a `Date` object → Firestore `Timestamp`).
2. Date-range query readers (these silently drop string-dated rows because timestamps sort before strings in Firestore):
   - `subscriptionUtils.ts:141` — `where("date", ">=", Timestamp.fromDate(subDays(new Date(), 365)))` → the Subscription Analyzer (`SubscriptionAnalyzer.tsx:71`) never sees manually-added transactions.
   - `cashflowUtils.ts:68` — `where("date", ">=", Timestamp.fromDate(startDate))` → Cash Flow dashboard misses the same rows.
   - `anomalyUtils.ts:114` — same pattern in `fetchTransactions`.
3. Client-side readers that treat the value as a Timestamp (`.toDate()` / `instanceof Timestamp`) get a `Timestamp` stringified into `"Timestamp(seconds=…, nanoseconds=…)"`, producing `Invalid Date`/`NaN` when the fallback path runs.

## Proposed Fix

- Canonicalize the writer: persist `date` uniformly (recommend storing an ISO string `'YYYY-MM-DD'` in all writers including `billUtils.markBillAsPaid`, or always a `Timestamp`).
- Add a data migration that normalizes existing mixed-type `date` fields (string ↔ Timestamp) before queries rely on them.
- Harden readers (`cashflowUtils`, `anomalyUtils`, `subscriptionUtils`, `trendsUtils`, `chatUtils`, `healthUtils`, `reportUtils`) to normalize `Timestamp`/`Date`/`string` to a single `Date` via one shared helper instead of ad-hoc `new Date(...)`.

ionfwsrijan:
## Problem

`CurrencyManager` wrote `date` as an ISO string while `billUtils` wrote a Firestore `Timestamp`. Each reader edge-parsed date shapes partially (some accepted strings, some only `Timestamp`/`Date`), so dates could be mis-read or silently fall back to "now".

## Fix

- `CurrencyManager` now persists `date: new Date(...)` on add and update, so manual transactions store a `Timestamp` just like bill transactions.
- Added a shared `toDate()` helper in `src/lib/utils.ts` that coerces `Timestamp` / `Date` / string / number consistently.
- Readers (`cashflowUtils`, `anomalyUtils`, `trendsUtils`, `reportUtils`, `chatUtils`) now parse dates through `toDate`, including both normal and `failed-precondition` fallback paths.

## Files changed

- `src/lib/utils.ts`
- `src/lib/cashflowUtils.ts`
- `src/lib/anomalyUtils.ts`
- `src/lib/trendsUtils.ts`
- `src/lib/reportUtils.ts`
- `src/lib/chatUtils.ts`

## Testing

- `npm test` passes.
- Scoped `tsc` typechecks pass; no new typecheck errors versus `origin/main`.

Closes #385

### DIFF
--- a/src/components/currency/CurrencyManager.tsx
+++ b/src/components/currency/CurrencyManager.tsx
@@ -33,6 +33,7 @@ import {
 } from 'lucide-react';
 import { toast } from 'sonner';
 import { motion } from 'motion/react';
+import { toDate } from '@/src/lib/utils';
 import {
   MAJOR_CURRENCIES,
   fetchExchangeRates,
@@ -64,7 +65,7 @@ interface Transaction {
   amount: number;
   currency: string;
   category: string;
-  date: string;
+  date: any;
   createdAt: any;
 }
 
@@ -198,7 +199,7 @@ export function CurrencyManager({ user }: CurrencyManagerProps) {
         amount: parseFloat(newTx.amount),
         currency: newTx.currency,
         category: newTx.category,
-        date: newTx.date,
+        date: new Date(newTx.date),
         createdAt: serverTimestamp(),
       });
       toast.success('Transaction added');
@@ -226,7 +227,7 @@ export function CurrencyManager({ user }: CurrencyManagerProps) {
         amount: parseFloat(editForm.amount),
         currency: editForm.currency,
         category: editForm.category,
-        date: editForm.date,
+        date: new Date(editForm.date),
       });
       toast.success('Transaction updated');
       setEditingTx(null);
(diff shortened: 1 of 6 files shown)


ionfwsrijan:

ionfwsrijan:

PR #401, opened 2026-08-05
fix: sells reduce holding quantity atomically and reject oversells

### LINKED ISSUE #386: Selling a holding never reduces its quantity - portfolio stays overstated
## Description

Recording a **sell** in the portfolio tracker writes a transaction record but never reduces the holding's `quantity` (or recomputes `avgCost`). Only the **buy** path touches the holding document. As a result, a holding you have fully sold keeps reporting its original quantity, and every portfolio metric — total value, P/L, allocation, top holdings — stays overstated.

## Expected Behaviour

A sell of `N` shares reduces the holding's `quantity` by `N`, recomputes `avgCost` (or tracks lots), and blocks selling more than owned. The portfolio UI reflects the sale immediately.

## Actual Behaviour

1. `portfolioUtils.ts:213-264` — `addTransaction`:
   - `buy` (`:229-256`): `addDoc` to `portfolioTransactions`, then updates the `portfolioHoldings` doc — `quantity += input.quantity`, new `avgCost` computed (`:247-254`).
   - every other type including `sell` (`:234-239`): only writes the transaction record via `setDoc`. There is **no** code path anywhere that decrements `holding.quantity` on a sell.
2. `fetchUserHoldings` (`portfolioUtils.ts:309-342`) just returns the stored `quantity` — unchanged after a sell.
3. `PortfolioTracker.tsx:113,218` loads holdings and transactions separately and never reconciles `sell` entries against holdings in the UI either.

So selling 100% of a position leaves the holding listed at full size forever, inflating `calculateTotalValue`, `calculateAllocation`, and `generatePortfolioSummary`. There is also no guard against selling more shares than are held.

## Proposed Fix

- In `addTransaction`, when `type === 'sell'`, atomically decrement the holding's `quantity` and recompute `avgCost` (or maintain lot-level data), using `increment()`/`runTransaction` to avoid races.
- Reject sells where `quantity > holding.quantity`.
- Decide how `dividend` / `deposit` / `withdrawal` should affect quantity/cost and implement accordingly, rather than leaving them write-only.
- Verify the fix in `PortfolioTracker` by refreshing holdings after a sale.

ionfwsrijan:
## Problem

Selling a holding never reduced the held quantity: `addTransaction` only handled `buy`, and there was no guard against selling more than you hold.

## Fix

- `portfolioUtils.addTransaction` now handles both `buy` and `sell`.
- A sell with no existing holding, or a sell exceeding the held quantity, is rejected with an error.
- A Firestore transaction re-validates the holding inside `runTransaction`, so an oversell cannot race a concurrent update.
- `avgCost` is only recalculated on buys; sells reduce `quantity` and keep the cost basis.
- `PortfolioTracker` surfaces the rejection reason in the error toast instead of a generic message.

## Files changed

- `src/lib/portfolioUtils.ts`
- `src/components/portfolio/PortfolioTracker.tsx`

## Testing

- `npm test` passes.

Closes #386

### DIFF
--- a/src/components/portfolio/PortfolioTracker.tsx
+++ b/src/components/portfolio/PortfolioTracker.tsx
@@ -229,7 +229,9 @@ export function PortfolioTracker({ user }: PortfolioTrackerProps) {
       }
     } catch (error) {
       handleFirestoreError(error, OperationType.CREATE, 'portfolioTransactions');
-      toast.error('Failed to add transaction');
+      toast.error(
+        error instanceof Error ? error.message : 'Failed to add transaction',
+      );
     }
   };
 
--- a/src/lib/portfolioUtils.ts
+++ b/src/lib/portfolioUtils.ts
@@ -15,6 +15,7 @@ import {
   addDoc,
   orderBy,
   deleteDoc,
+  runTransaction,
 } from 'firebase/firestore';
 import { db, handleFirestoreError, OperationType } from './firebase';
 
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #402, opened 2026-08-05
fix: exclude current month from balance projection double-count

### LINKED ISSUE #387: Cash-flow balance projection double-counts the current month
## Description

The Cash Flow dashboard's balance projection **double-counts the current month**. `calculateBalanceProjection` seeds its running balance with *all* fetched transactions (including the current month's actuals), then iterates a forecast whose first month is the **current month** with a projected net that is itself derived from the current month's transactions. The current month therefore appears in the balance line twice — once as actuals, once as a projection.

## Expected Behaviour

The projected balance should start from today's actual balance and project **forward** from next month, so no month is counted twice.

## Actual Behaviour

1. `cashflowUtils.ts:46-54` — `getNextMonths(6)` starts at `i = 0`, so the first forecast month is the **current** month.
2. `cashflowUtils.ts:136-186` — `calculateMonthlyForecast` buckets the fetched transactions (which include the current, partial month) and averages per month; the first projected month is the current month and its `projectedNet` includes current-month actuals.
3. `cashflowUtils.ts:188-205` — `calculateBalanceProjection`:
   - `:192-196` — `currentBalance` sums **all** transactions, including the current month's income/expenses.
   - `:198-204` — then adds `f.projectedNet` for each forecast month, and the first iteration is the current month again.
4. Wired in `CashFlowDashboard.tsx:65-67` (`fetchUserTransactions(6)` → `calculateMonthlyForecast` → `calculateBalanceProjection`).

Result: today's projected balance is inflated by roughly one extra month of net income/expense, and the projection chart's first point misrepresents the current state.

## Proposed Fix

- Start projections at the **next** month: in `getNextMonths`, begin at `i = 1` (or slice the current month out of the forecast used by the projection).
- Optionally seed `currentBalance` from transactions **before** the current month and let the forecast advance it, keeping current-month actuals out of the additive loop.
- Add a unit test asserting `projectedBalance[0]` equals today's real balance, not today's balance plus a projected month.

ionfwsrijan:
## Problem

The balance projection seeded `currentBalance` from all actual transactions (which already include the current, partial month) and then added the current month's `projectedNet` as the first forecast point. The current month was effectively counted twice, inflating the projected balance.

## Fix

- `calculateBalanceProjection` now reports the seeded actual balance for the current month without adding the current month's `projectedNet`.
- Only future months advance the balance, so no month is double-counted.

## Files changed

- `src/lib/cashflowUtils.ts`
- `tests/cashflow-balance-projection.test.mjs` (new)

## Testing

- New unit tests assert that the projection seeds from the real transaction balance, reports the current month without adding its projected net, and only advances on future months.

Closes #387

### DIFF
--- a/src/lib/cashflowUtils.ts
+++ b/src/lib/cashflowUtils.ts
@@ -189,14 +189,23 @@ export function calculateBalanceProjection(
   transactions: Transaction[],
   forecast: ForecastData[],
 ): BalanceProjection[] {
+  // Seed with the sum of all fetched transactions: today's real balance,
+  // which already includes the current (partial) month's activity.
   let currentBalance = 0;
   transactions.forEach((t) => {
     if (t.type === "income") currentBalance += t.amount;
     else currentBalance -= t.amount;
   });
 
+  // The current month's actuals are already folded into currentBalance, so
+  // adding its projected net on top would count the month twice. The current
+  // month reports the real balance; only future months advance the balance.
+  const currentMonth = getMonthKey(new Date());
+
   return forecast.map((f) => {
-    currentBalance += f.projectedNet;
+    if (f.month !== currentMonth) {
+      currentBalance += f.projectedNet;
+    }
     return {
       month: f.month,
       projectedBalance: Math.round(currentBalance * 100) / 100,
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

AakashRathore136:
@copilot resolve the merge conflicts in this pull request

ionfwsrijan:

PR #403, opened 2026-08-05
fix: hold concurrency slot for the full analysis pipeline lifetime

### LINKED ISSUE #388: Analysis concurrency slot is freed on client disconnect while the pipeline keeps running
## Description

The in-flight analysis concurrency limiter releases a user's slot as soon as the HTTP response stream closes — including when the **client disconnects** mid-analysis. The pipeline (PDF parsing → Hugging Face inference → Storage upload → Firestore writes) keeps running long after that, so the freed slot no longer corresponds to the work that is actually in progress. This lets a client abort requests to escape the "max 2 concurrent analyses" guard and run unlimited concurrent pipelines, while the server still performs the expensive AI work.

## Expected Behaviour

A concurrency slot must be held for the full lifetime of the analysis pipeline and only released when that pipeline actually settles. A disconnected client should not refund the slot while its analysis continues to run (and ideally the pipeline should be cancelled on disconnect).

## Actual Behaviour

`server.ts:539-570` — `concurrentAnalyzeLimiter`:

1. `:543-556` — increments `inFlightAnalyzeByUser` and passes the request through when below the cap of 2.
2. `:558-565` — registers a single `cleanup` that decrements the counter.
3. `:567-568` — cleanup is bound to **both** `res.once("finish")` and `res.once("close")`.
4. `close` fires on an **aborted/closed client connection**, not just on normal completion. The handler (`:578+`) continues its async work independently of the response object and never checks `res.writableEnded`.

Concrete failure: a free user fires 2 analyses, then submits a 3rd → correctly `429`. They abort the 2 in-flight requests (navigate away / cancel in the browser). Both slots free immediately while the pipelines keep executing, and a 3rd/4th/5th request is now accepted — the effective concurrency exceeds 2, and the HF inference + storage load runs unbounded relative to the guard. Conversely, if a legitimate user's connection drops, the limiter loses track of the running job entirely.

## Proposed Fix

- Release the slot only when the async pipeline handler settles: move cleanup into the handler's `finally` (or use a per-request "done" signal) instead of relying on `res.once("close")`.
- On client disconnect, either abort the pipeline (pass an `AbortSignal` derived from `req.on('close')` into the inference/storage steps) or keep the slot held until the work completes so the counter stays accurate.
- Consider making the counter keyed off the actual pipeline execution rather than request lifecycle.

ionfwsrijan:
## Problem

The in-flight analysis concurrency limiter released a user's slot via `res.once("close")`. `close` fires on a client disconnect while the analysis pipeline (PDF parsing -> inference -> Storage upload -> Firestore writes) keeps running, so the freed slot no longer matched the work actually in progress. Clients could abort requests to escape the "max 2 concurrent analyses" guard while the server continued running expensive pipelines.

## Fix

- `concurrentAnalyzeLimiter` no longer binds cleanup to the `close` event.
- The release function is exposed via `res.locals.releaseAnalyzeSlot` and called from the analyze handler's `finally`, so the slot is held for the full pipeline lifetime and only released once the pipeline settles.
- `res.once("finish")` is kept as a safety net only for requests rejected before the handler body runs (e.g. an oversized upload rejected by `upload.single`), which fires after the response is actually sent.

## Files changed

- `server.ts`

## Testing

- `npm test` passes.
- Scoped `tsc` typecheck of `server.ts` passes.

Closes #388

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -564,8 +564,16 @@ async function startServer() {
       else inFlightAnalyzeByUser.set(key, updated);
     };
 
+    // The slot must be held for the full lifetime of the analysis pipeline.
+    // Binding cleanup to `res.once("close")` would refund it as soon as the
+    // client disconnects, even though the pipeline keeps running afterwards,
+    // letting clients escape the "max 2 concurrent analyses" guard. The
+    // handler releases the slot in its `finally` once the pipeline settles;
+    // the `finish` binding below is only a safety net for requests that never
+    // reach the handler body (e.g. an oversized upload rejected by
+    // `upload.single`), which fires after the (error) response is sent.
+    res.locals.releaseAnalyzeSlot = cleanup;
     res.once("finish", cleanup);
-    res.once("close", cleanup);
     next();
   };
 
@@ -1028,6 +1036,11 @@ CRITICAL RULES:
         }
 
         return res.status(500).json(errorResponse);
+      } finally {
+        // Release the concurrency slot only once the pipeline has settled.
+        // It must not be freed while the pipeline continues to run, e.g. after
+        // a client disconnects mid-analysis.
+        res.locals?.releaseAnalyzeSlot?.();
       }
     },
   );


ionfwsrijan:

PR #404, opened 2026-08-05
fix: concurrency rejections no longer consume daily analysis quota

### LINKED ISSUE #389: Concurrent-limit 429s burn the 5/day analysis quota - free users locked out for 24h
## Description

The daily analysis quota (5 per user per day) is consumed by requests that are **rejected** by the concurrent-analysis limiter. Because `analyzeRateLimiter` runs *before* `concurrentAnalyzeLimiter`, every extra request that gets a `429` from the concurrency gate has already incremented the daily counter. A free user whose first analysis is slow — who retries a couple of times while the two concurrent slots are occupied — can burn the entire 5/day budget without a single successful analysis, then be locked out for 24 hours.

## Expected Behaviour

Only requests that actually start an analysis (i.e. pass the concurrency gate) should count against the daily quota. A `429 CONCURRENT_LIMIT` rejection must not also consume one of the 5 daily attempts.

## Actual Behaviour

`server.ts`:

1. `:517-534` — `analyzeRateLimiter` (`windowMs: 24h`, `max: 5`) is registered first.
2. `:539-570` — `concurrentAnalyzeLimiter` (max 2 in flight) is registered second.
3. `:573-577` — `app.post('/api/analyze', analyzeRateLimiter, concurrentAnalyzeLimiter, ...)` — so the quota middleware runs first on every request.
4. `express-rate-limit` increments the key counter for every request that reaches the middleware, whether it later passes or is rejected. Requests rejected at the concurrency stage have already been counted.

Example: user submits an analysis (1/5 used), then a second (2/5) and third (3/5) — the third gets `429 CONCURRENT_LIMIT` but still counts. Two more tries → `429` again, now 5/5 used. The user now has **zero** remaining analyses for 24h even though only one analysis actually ran. This turns a transient concurrency rejection into a permanent-feeling outage for free-tier users.

## Proposed Fix

- Reorder the middlewares so `concurrentAnalyzeLimiter` runs **before** `analyzeRateLimiter`, so rejected requests never reach the quota counter.
- Or: only count a request once it has passed the concurrency gate (e.g., move the rate-limit `incr` into the handler / use a limiter whose `keyGenerator`+skip logic excludes concurrency-rejected requests).
- Add a test that fires 3 simultaneous requests with 2 in flight and asserts the 3rd (429) does not reduce the remaining daily quota.

ionfwsrijan:
## Problem

`analyzeRateLimiter` (5/day quota) ran before `concurrentAnalyzeLimiter`, and `express-rate-limit` counts every request that reaches it. Every extra request rejected with `429 CONCURRENT_LIMIT` still consumed one of the 5 daily attempts, so a free user whose first analysis was slow could burn the whole daily budget on rejected requests and be locked out for 24 hours.

## Fix

- Reversed the middleware order on `POST /api/analyze`: `concurrentAnalyzeLimiter` now runs before `analyzeRateLimiter`, so requests rejected at the concurrency gate never reach the quota counter.
- Added a guard test asserting the ordering so it cannot silently regress.

## Files changed

- `server.ts`
- `tests/analyze-quota-guard.test.mjs` (new)

## Testing

- New test verifies the concurrency limiter is registered before the quota limiter on `/api/analyze`.

Closes #389

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -570,10 +570,13 @@ async function startServer() {
   };
 
   // AI Analysis Endpoint
+  // `concurrentAnalyzeLimiter` must run BEFORE `analyzeRateLimiter`: the rate
+  // limiter counts every request that reaches it, so a request rejected at the
+  // concurrency gate (429 CONCURRENT_LIMIT) must never consume daily quota.
   app.post(
     "/api/analyze",
-    analyzeRateLimiter,
     concurrentAnalyzeLimiter,
+    analyzeRateLimiter,
     upload.single("file"),
     async (req: any, res) => {
       try {
--- a/tests/analyze-quota-guard.test.mjs
+++ b/tests/analyze-quota-guard.test.mjs
@@ -1,0 +1,44 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+
+const repoRoot = path.resolve(path.dirname(fileURLToPath(import.meta.url)), "..");
+const server = readFileSync(path.join(repoRoot, "server.ts"), "utf8").replace(
+  /\r\n/g,
+  "\n",
+);
+
+// /api/analyze middleware must reject at the concurrency gate before the
+// daily-quota counter is ever touched. express-rate-limit counts every request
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #405, opened 2026-08-05
fix: apply goal contributions atomically with server-side increment

### LINKED ISSUE #390: Rapid goal contributions silently overwrite each other (stale read-then-write)
## Description

Adding contributions to a financial goal uses a stale read-then-write: `GoalPlanner` reads `currentAmount` from React state (which only refreshes when the Firestore snapshot round-trips) and writes back an **absolute** new amount. Two rapid "Add" clicks — or an add racing the auto-complete write — both compute from the same old value, so the second write overwrites the first and a real contribution is silently lost.

## Expected Behaviour

Every contribution is applied on top of the latest persisted `currentAmount`, so rapid or concurrent adds never lose money.

## Actual Behaviour

1. `GoalPlanner.tsx:215-230` — `updateGoalAmount`:
   - `:216` — `const goal = goals.find((g) => g.id === goalId)` (from React state).
   - `:220` — `const newAmount = goal.currentAmount + addedAmount;` then `:221-223` — `updateDoc(doc(db, 'goals', goalId), { currentAmount: newAmount })` — an absolute value derived from stale state.
2. `GoalCard.tsx:49-54` — `handleAddContribution` calls `onUpdateAmount(goal.id, amount)` directly from the input; nothing disables the button while a write is in flight, so two quick clicks fire two concurrent updates from the same snapshot.
3. The same stale-state pattern exists in the auto-complete path (`GoalPlanner.tsx:132-151`), which writes `completedAmount`/`currentAmount` while a manual add may be in flight.

Failure path: goal has `currentAmount = 100`. User clicks "Add 50" twice quickly. Both handlers read `100`, both write `150`. One 50-unit contribution vanishes. If the second click races the auto-complete effect (`status === 'completed'`), the mismatch can also leave a goal marked completed at the wrong amount.

## Proposed Fix

- Use a Firestore server-side `increment(addedAmount)` (`{ currentAmount: increment(addedAmount) }`) in `updateGoalAmount` so the write is atomic and race-free.
- Alternatively, wrap the read-modify-write in `runTransaction`.
- In `GoalCard`, disable the add button (or queue) while a contribution write is pending; in `GoalPlanner`, base the auto-complete write on the same atomic increment rather than a snapshot value.

ionfwsrijan:
## Problem

Goal contributions used a stale read-then-write: `GoalPlanner` read `currentAmount` from React state and wrote back an absolute new amount. Two rapid "Add" clicks, or an add racing the auto-complete write, both computed from the same old value, so the second write overwrote the first and a real contribution was silently lost.

## Fix

- `updateGoalAmount` now applies `currentAmount: increment(addedAmount)` atomically on the server, so rapid or concurrent adds can never overwrite each other.
- The goal is added to `updatingGoalIds` while the write is in flight so the auto-complete effect cannot write a stale snapshot amount on top of it.
- `GoalCard` disables the input and button while a contribution write is pending and only clears the input on success.
- Removed a duplicate `GoalCard` import in `GoalPlanner`.

## Files changed

- `src/components/goals/GoalPlanner.tsx`
- `src/components/goals/GoalCard.tsx`

## Testing

- `npm test` passes.

Closes #390

### DIFF
--- a/src/components/goals/GoalCard.tsx
+++ b/src/components/goals/GoalCard.tsx
@@ -23,7 +23,7 @@ import {
 
 interface GoalCardProps {
   goal: Goal;
-  onUpdateAmount: (goalId: string, amount: number) => void;
+  onUpdateAmount: (goalId: string, amount: number) => Promise<void> | void;
   onViewDetails: (goal: Goal) => void;
   onStatusChange: (goalId: string, status: Goal['status']) => void;
   onDelete: (goalId: string) => void;
@@ -37,6 +37,7 @@ export function GoalCard({
   onDelete,
 }: GoalCardProps) {
   const [amountInput, setAmountInput] = useState('');
+  const [adding, setAdding] = useState(false);
   const daysRemaining = calculateDaysRemaining(goal.deadline);
   const progress = getProgressPercentage(goal.currentAmount, goal.targetAmount);
   const monthlyContribution = calculateMonthlyContribution(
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

PR #406, opened 2026-08-05
fix: use leave-one-out baseline for large-transaction anomaly detection

### LINKED ISSUE #391: Anomaly detection uses a self-inclusive baseline - large transactions are never flagged
## Description

The anomaly detector computes each transaction's outlier threshold from a baseline that **includes the transaction itself**. `detectLargeTransactions` compares every amount against `mean + 2*stdDev` where `mean`/`stdDev` were computed over the full category set including the very amount being tested. Large transactions pull their own baseline up, so in small categories they rarely exceed the threshold and the anomaly engine systematically misses the biggest anomalies it exists to find.

## Expected Behaviour

A transaction should be tested against a baseline that excludes it (leave-one-out), or against a prior-period baseline, so a genuinely large amount stands out.

## Actual Behaviour

1. `anomalyUtils.ts:150-189` — `calculateCategoryBaseline` computes `mean` and `stdDev` over **all** transactions in the category (`:161-171`), including future/current outliers.
2. `anomalyUtils.ts:191-199` — `detectLargeTransactions`:
   - `:198` — `const threshold = categoryBaseline.mean + categoryBaseline.stdDev * 2;`
   - `:199` — `return Math.abs(transaction.amount) > Math.max(threshold, 1000);` — the candidate is inside `mean`/`stdDev`.
3. Wired in `AnomalyDashboard.tsx:351-355` — `fetchUserTransactions(userId, 3)` → `calculateCategoryBaseline(transactions)` → `detectLargeTransactions(transactions, baseline)`.

Concrete failure: a category has 3 transactions — 100, 100, 10,000. `mean = 3400`, `stdDev ≈ 4671`, threshold `≈ 12,742`. The 10,000 transaction is **not** flagged (10,000 < 12,742), even though it dominates the category. For any category with a single dominant expense, self-inclusion inflates the threshold and suppresses the alert; the `Math.max(…, 1000)` floor only makes this worse by hiding medium categories entirely.

## Proposed Fix

- Test each transaction against a leave-one-out baseline: for each candidate, recompute `mean`/`stdDev` over the category **excluding** it (or compute baseline statistics once over a rolling/prior window).
- Enforce a minimum sample size per category (e.g., ≥ 4 observations) before flagging, and document that.
- Add unit tests for `calculateCategoryBaseline` + `detectLargeTransactions` covering the single-dominant-transaction case.

ionfwsrijan:
## Problem

`detectLargeTransactions` compared every amount against `mean + 2*stdDev`, where `mean`/`stdDev` were computed over the full category set including the very amount being tested. Large transactions pulled their own baseline up, so in small categories they rarely exceeded the threshold and the engine missed the biggest anomalies it exists to find.

## Fix

- `detectLargeTransactions` now tests each candidate against a leave-one-out baseline: `mean`/`stdDev` are recomputed over the other transactions in the same category, so a large amount cannot inflate its own threshold.
- Enforced a minimum sample size of 4 observations per category before flagging, documented via `MIN_LARGE_TRANSACTION_SAMPLES`.
- Extracted the pure statistics (`calculateCategoryBaseline`, `detectLargeTransactions`, and their types) into `src/lib/anomalyStats.ts`, which has no firebase dependency and can be unit tested under Node. `anomalyUtils.ts` re-exports the same API, so no call sites change.

## Files changed

- `src/lib/anomalyStats.ts` (new)
- `src/lib/anomalyUtils.ts`
- `tests/anomaly-leave-one-out.test.mjs` (new)

## Testing

- New unit tests cover `calculateCategoryBaseline` aggregation and the single-dominant-transaction case, and verify categories below the minimum sample size are not flagged.
- `npm test` passes; no new typecheck errors versus `origin/main`.

Closes #391

### DIFF
--- a/src/lib/anomalyStats.ts
+++ b/src/lib/anomalyStats.ts
@@ -1,0 +1,108 @@
+/**
+ * @license
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { format } from "date-fns";
+
+export interface Transaction {
+  id: string;
+  userId: string;
+  amount: number;
+  category: string;
+  date: Date;
+  description?: string;
+  type?: "expense" | "income";
+}
+
+export interface CategoryBaseline {
+  mean: number;
+  stdDev: number;
+  monthlyTotals: number[];
+}
+
+// Minimum number of observations per category before large-transaction
+// anomalies are flagged. Below this the leave-one-out baseline has too few
+// comparison points to be statistically meaningful.
+export const MIN_LARGE_TRANSACTION_SAMPLES = 4;
+
+export function calculateCategoryBaseline(
+  transactions: Transaction[],
+): Map<string, CategoryBaseline> {
+  const grouped = new Map<string, Transaction[]>();
+
+  transactions.forEach((transaction) => {
+    const category = transaction.category || "Other";
+    grouped.set(category, [...(grouped.get(category) || []), transaction]);
+  });
+
+  const baseline = new Map<string, CategoryBaseline>();
+  grouped.forEach((items, category) => {
+    const amounts = items.map((item) => Math.abs(item.amount));
+    const mean =
+      amounts.length > 0
+        ? amounts.reduce((sum, amount) => sum + amount, 0) / amounts.length
+        : 0;
+    const variance =
+      amounts.length > 0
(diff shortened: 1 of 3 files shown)


ionfwsrijan:

PR #436, opened 2026-08-05
fix: Docker build copies /app/dist and runs dist/server.cjs

### LINKED ISSUE #426: [Bug] Docker image build always fails: Dockerfile copies /app/build and runs build/server.cjs but npm run build emits dist/server.cjs
**Description**

The Dockerfile references a build artifact that `npm run build` never produces:

- `Dockerfile:22` — `COPY --from=builder --chown=node:node /app/build ./build`
- `Dockerfile:28` — `CMD ["node", "build/server.cjs"]`
- `package.json:8` — `"build": "vite build && esbuild server.ts --bundle --platform=node --format=cjs --packages=external --sourcemap --outfile=dist/server.cjs"`

The build script emits the server bundle to `dist/server.cjs`, and neither `vite build` nor esbuild ever creates a `build/` directory.

**Expected Behavior**

`docker build` completes and the container starts the API server.

**Actual Behavior**

- `COPY --from=builder /app/build ./build` fails because `/app/build` does not exist in the builder stage, so `docker build` aborts.
- Even if the copy were fixed, `CMD ["node", "build/server.cjs"]` points at a non-existent file, so the container would crash on boot. Note the repo's own `start` script (`package.json:9`) correctly runs `node dist/server.cjs`.

**Proposed Fix**

- Change `Dockerfile:22` to `COPY --from=builder --chown=node:node /app/dist ./dist`.
- Change `Dockerfile:28` to `CMD ["node", "dist/server.cjs"]`.
- (Optional) Add a container smoke test to CI that runs `docker build` to keep the two in sync.

ionfwsrijan:
## Problem

The Dockerfile references a build artifact that `npm run build` never produces:

- `Dockerfile` — `COPY --from=builder --chown=node:node /app/build ./build`
- `Dockerfile` — `CMD ["node", "build/server.cjs"]`
- `package.json` — `"build": "vite build && esbuild server.ts --bundle --platform=node --format=cjs --packages=external --sourcemap --outfile=dist/server.cjs"`

The build script emits the server bundle to `dist/server.cjs` (and `vite build` writes to `dist/`). Neither step ever creates a `build/` directory, so `docker build` aborts at the `COPY /app/build` step, and even if it got past that, the container would crash on boot because `build/server.cjs` does not exist.

## Fix

- Removed the invalid `COPY --from=builder --chown=node:node /app/build ./build` line (the `/app/dist` copy was already present and correct).
- Changed the runtime `CMD` to `node dist/server.cjs`, matching the repo's own `start` script (`node dist/server.cjs`).

## Files changed

- `Dockerfile` — removed the stale `/app/build` copy and pointed `CMD` at `dist/server.cjs`.

## Testing

- The two referenced paths now match what `npm run build` emits (`dist/server.cjs`), so the builder stage always contains the copied artifact and the container starts the API server (port `3001`, matching `server.ts`).

Closes #426

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,10 +19,9 @@ ENV NODE_ENV=production
 COPY --from=builder --chown=node:node /app/package*.json ./
 COPY --from=builder --chown=node:node /app/node_modules ./node_modules
 COPY --from=builder --chown=node:node /app/dist ./dist
-COPY --from=builder --chown=node:node /app/build ./build
 
 USER node
 
 EXPOSE 3001
 
-CMD ["node", "build/server.cjs"]
+CMD ["node", "dist/server.cjs"]


ionfwsrijan:

PR #437, opened 2026-08-05
fix: persist failed analysis records that match Firestore rules

### LINKED ISSUE #427: [Bug] Failed-analysis error record is always rejected by Firestore rules: FileUpload writes to /analyses without documentId and riskLevel required by isValidAnalysis
**Description**

When an AI analysis pipeline run fails, `FileUpload.tsx` tries to persist a `status: "failed"` record to the top-level `analyses` collection, but the write is always rejected by the Firestore security rules, so failed runs are never recorded.

- `src/components/FileUpload.tsx:190-199` writes:
  ```ts
  await addDoc(collection(db, "analyses"), {
    fileName: file?.name || "Unknown",
    fileSize: file?.size || 0,
    status: "failed",
    errorMessage: userMsg,
    errorKind: kind,
    uploadedAt: serverTimestamp(),
    failedAt: serverTimestamp(),
    ownerId: user?.uid || null,
  });
  ```
- `firestore.rules:35-39` requires `isValidAnalysis`, which demands `data.keys().hasAll(['documentId', 'ownerId', 'riskLevel'])` and `ownerId == request.auth.uid`.
- `firestore.rules:97-104` gates `create` on `/analyses/{analysisId}` with `isSignedIn() && isValidAnalysis(incoming())`.

The failure record omits `documentId` and `riskLevel`, and sets `ownerId: null` when signed out, so `isValidAnalysis` fails. `isSignedIn()` also requires `email_verified`. The permission-denied error is swallowed at `FileUpload.tsx:200-202` ("Could not persist failed record").

**Expected Behavior**

Failed pipeline runs (auth errors, quota, invalid PDFs, server errors) are persisted as `status: "failed"` records so the client can surface diagnostics and history.

**Actual Behavior**

Every failed run throws `permission-denied` from the client SDK and the failure record is never written. The error-persistence feature is dead code; diagnostics in the error panel can never be reconstructed from Firestore.

**Proposed Fix**

- Either extend the failure record with `documentId`/`riskLevel` and a non-null `ownerId` (drop the null fallback and bail out when signed out), or
- Add a separate rules branch for `/analyses/{analysisId}` allowing `create` of `status == 'failed'` records with a validation function that matches the actual document shape, so the rules and the write contract agree.

ionfwsrijan:
## Problem

When an AI analysis pipeline run fails, `FileUpload.tsx` tries to persist a `status: "failed"` record to the top-level `analyses` collection, but the write is always rejected by the Firestore security rules, so failed runs are never recorded:

- The failure record omits `documentId` and `riskLevel` (fields required by `isValidAnalysis`), and falls back to `ownerId: null` when signed out.
- The `create` rule on `/analyses/{analysisId}` gates every write behind `isValidAnalysis(incoming())`, so the record never matches and the client gets `permission-denied`.
- The resulting error is swallowed, so the error-persistence feature is effectively dead code.

## Fix

Made the write contract and the Firestore rules agree on the shape of a failed-analysis record:

- `src/components/FileUpload.tsx` — bail out of persisting when there is no signed-in user (a record without an owner can never be stored), and always write a non-null `ownerId: user.uid`.
- `firestore.rules` — added `isValidFailedAnalysis()` which validates the actual failed-record shape (`status == 'failed'` owned by the requester), and allowed `create` on `/analyses/{analysisId}` for either a valid analysis or a valid failed analysis.

Failed records remain scoped per-user: they are only writable by the signed-in owner and readable by the owner (or an admin).

## Files changed

- `src/components/FileUpload.tsx` — non-null `ownerId`, skip persistence when signed out.
- `firestore.rules` — new `isValidFailedAnalysis` helper and updated `create` rule for `/analyses/{analysisId}`.

## Testing

- `npx tsc --noEmit` reports no new errors for `FileUpload.tsx` (remaining errors are pre-existing on `main` in `RolloverManager.tsx`).
- Rules now match the exact document shape written by the failure handler, so the write passes the `create` gate when the user is signed in and email-verified.

Closes #427

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -38,6 +38,12 @@ service cloud.firestore {
              data.riskLevel in ['low', 'medium', 'high'];
     }
 
+    function isValidFailedAnalysis(data) {
+      return data.keys().hasAll(['status', 'ownerId']) &&
+             data.status == 'failed' &&
+             data.ownerId == request.auth.uid;
+    }
+
     // ----------------------------------------------------
     // COLLECTION MATCHES
     // ----------------------------------------------------
@@ -98,7 +104,7 @@ service cloud.firestore {
       // SECURITY: Per-user RLS for any top-level analyses collection
       // Prevents unauthorized access to sensitive financial analysis data
       allow read: if isOwner(existing().ownerId) || isAdmin();
-      allow create: if isSignedIn() && isValidAnalysis(incoming());
+      allow create: if isSignedIn() && (isValidAnalysis(incoming()) || isValidFailedAnalysis(incoming()));
       allow update: if isAdmin();
       allow delete: if isOwner(existing().ownerId) || isAdmin();
     }
--- a/src/components/FileUpload.tsx
+++ b/src/components/FileUpload.tsx
@@ -184,6 +184,7 @@ export function FileUpload({ user, onComplete, onCancel }: any) {
       toast.error(userMsg);
 
       try {
+        if (!user?.uid) return;
         const { db } = await import("@/src/lib/firebase");
         const { collection, addDoc, serverTimestamp } =
           await import("firebase/firestore");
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #439, opened 2026-08-05
fix: restore rejected budget categories and their status after reload

### LINKED ISSUE #429: [Bug] Rejected budget categories silently reappear as accepted after reload: saved 0 amount treated as falsy by categoryBudgets lookup
**Description**

Rejected budget categories silently reappear as "accepted" after a reload because a saved amount of `0` is treated as falsy by the budget restore logic.

- `src/components/budget/BudgetDashboard.tsx:158-163` — `handleSaveBudget` persists a rejected category as `categoryBudgets[cat] = 0`:
  ```ts
  const amount = s.status === "rejected" ? 0 : (s.modifiedAmount ?? s.suggestedAmount);
  categoryBudgets[s.category] = amount;
  ```
- `src/components/budget/BudgetDashboard.tsx:126-133` — on load, the saved value is read back with `||` fallback and the status is hardcoded to `accepted`:
  ```ts
  suggestedAmount: savedBudget.categoryBudgets[s.category] || s.suggestedAmount,
  modifiedAmount: savedBudget.categoryBudgets[s.category] || s.suggestedAmount,
  status: "accepted" as const,
  ```

Because `0` is falsy, the saved `0` is discarded and a freshly generated suggestion is substituted; the category is forced to `status: "accepted"`. Meanwhile `finalTotal` is set from `savedBudget.totalBudget` (which was computed excluding rejected categories), so the visible category list no longer adds up to the displayed total.

**Expected Behavior**

A category the user explicitly rejected stays rejected (or hidden) after reload, and the category list reconciles with `totalBudget`.

**Actual Behavior**

After saving a budget with rejected categories and reloading, every rejected category comes back as accepted with a new auto-suggested amount. The user's explicit decision is silently undone, and the per-category amounts disagree with the loaded `totalBudget`.

**Proposed Fix**

- In the restore path, check key presence instead of truthiness, e.g.:
  ```ts
  Object.prototype.hasOwnProperty.call(savedBudget.categoryBudgets, s.category)
    ? savedBudget.categoryBudgets[s.category]
    : s.suggestedAmount
  ```
- Persist and restore the per-category `status` (not just `0`) so rejected categories remain rejected.

ionfwsrijan:
## Problem

Rejected budget categories silently reappear as "accepted" after a reload because a saved amount of `0` is treated as falsy by the budget restore logic:

- `handleSaveBudget` persists a rejected category as `categoryBudgets[cat] = 0`.
- On load, the saved value is read back with a `||` fallback (`savedBudget.categoryBudgets[s.category] || s.suggestedAmount`) and the status is hardcoded to `accepted`.

Because `0` is falsy, the saved `0` is discarded and a freshly generated suggestion is substituted, and the category is forced to `status: "accepted"`. The user's explicit rejection is silently undone, and the per-category amounts disagree with the loaded `totalBudget`.

## Fix

- **Restore path:** checks key presence with `Object.prototype.hasOwnProperty` instead of truthiness, so a saved `0` is used. The per-category `status` is now restored instead of being hardcoded to `accepted`, so rejected categories remain rejected.
- **Save path:** persists the per-category `status` alongside the amounts via a new `categoryStatuses` field on the budget document, so the restore logic has the actual status to work with.

With rejected categories restored as rejected, `calculateTotalBudget`-based `totalBudget` (which excludes rejected categories) reconciles with the visible category list again.

## Files changed

- `src/components/budget/BudgetDashboard.tsx` — key-presence check in the restore path, restore the saved per-category status, and persist `categoryStatuses` on save.

## Testing

- `npx tsc --noEmit` reports no new errors for `BudgetDashboard.tsx` (remaining errors are pre-existing on `main`).
- Verified the restore logic keeps a saved `0` (rejected) category and its `rejected` status after reload instead of substituting a new suggestion.

Closes #429

### DIFF
--- a/src/components/budget/BudgetDashboard.tsx
+++ b/src/components/budget/BudgetDashboard.tsx
@@ -123,14 +123,22 @@ export function BudgetDashboard({ user }: { user: any }) {
       let finalConfidence = confidence;
 
       if (savedBudget && savedBudget.categoryBudgets) {
-        finalSuggestions = generatedSuggestions.map((s) => ({
-          ...s,
-          suggestedAmount:
-            savedBudget.categoryBudgets[s.category] || s.suggestedAmount,
-          modifiedAmount:
-            savedBudget.categoryBudgets[s.category] || s.suggestedAmount,
-          status: "accepted" as const,
-        }));
+        finalSuggestions = generatedSuggestions.map((s) => {
+          const hasSaved = Object.prototype.hasOwnProperty.call(
+            savedBudget.categoryBudgets,
+            s.category,
+          );
+          const savedAmount = hasSaved
+            ? savedBudget.categoryBudgets[s.category]
+            : s.suggestedAmount;
+          const savedStatus = savedBudget.categoryStatuses?.[s.category];
+          return {
+            ...s,
+            suggestedAmount: savedAmount,
+            modifiedAmount: savedAmount,
+            status: savedStatus || (hasSaved ? "accepted" : s.status),
+          };
+        });
         finalTotal = savedBudget.totalBudget;
         finalConfidence = savedBudget.confidenceScore;
         setSaved(true);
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #440, opened 2026-08-05
fix: use latest snapshot for portfolio All-Time Return card

### LINKED ISSUE #430: [Bug] Portfolio All-Time Return card shows the oldest snapshot P/L instead of the latest (index confusion in desc-ordered history)
**Description**

The portfolio "All-Time Return" card displays the P/L of the OLDEST portfolio snapshot instead of the latest one, because the history is fetched in descending order but the card reads the last array element.

- `src/lib/portfolioUtils.ts:553-587` — `fetchPortfolioHistory` orders by `orderBy('snapshotDate', 'desc')`, so index `0` is the newest snapshot and `[length-1]` is the oldest.
- `src/components/portfolio/PortfolioTracker.tsx:666-679` — the "All-Time Return" card reads `performanceHistory[performanceHistory.length - 1]` (oldest) for `profitLoss` and `profitLossPercent`.
- `src/components/portfolio/PortfolioTracker.tsx:686` — the adjacent "Latest Total Value" card correctly reads `performanceHistory[0]` (newest).

**Expected Behavior**

"All-Time Return" reflects the latest cumulative return (the newest snapshot), consistent with the "Latest Total Value" card and the live header P/L.

**Actual Behavior**

"All-Time Return" is frozen at the earliest snapshot's P/L. It ignores all later price changes, so it can disagree with the current header P/L and the "Latest Total Value" card in the same row. In a financial app this shows a stale, misleading return figure.

**Proposed Fix**

- Read `performanceHistory[0]` for the return card (and its percentage), matching the "Latest Total Value" card, e.g. `performanceHistory[0]?.profitLoss` / `performanceHistory[0]?.profitLossPercent`.
- Keep the line chart's ascending reversal as-is, but centralize the "latest snapshot" indexing so the two cards cannot drift apart again.

ionfwsrijan:
## Problem

The portfolio "All-Time Return" card displays the P/L of the OLDEST portfolio snapshot instead of the latest one:

- `fetchPortfolioHistory` (`src/lib/portfolioUtils.ts`) orders by `orderBy('snapshotDate', 'desc')`, so index `0` is the newest snapshot and `[length - 1]` is the oldest.
- `PortfolioTracker.tsx` read `performanceHistory[performanceHistory.length - 1]` (oldest) for the "All-Time Return" `profitLoss` / `profitLossPercent`, while the adjacent "Latest Total Value" card correctly reads `performanceHistory[0]` (newest).

The result: "All-Time Return" is frozen at the earliest snapshot's P/L, ignores later price changes, and can disagree with the current header P/L and the "Latest Total Value" card in the same row.

## Fix

- "All-Time Return" now reads `performanceHistory[0]` for both the dollar amount and the percentage, matching the "Latest Total Value" card, so both cards reflect the same latest snapshot.
- The line chart's ascending reversal (`[...performanceHistory].reverse()`) is kept as-is.

## Files changed

- `src/components/portfolio/PortfolioTracker.tsx` — All-Time Return card now indexes `performanceHistory[0]` (latest snapshot) instead of `[length - 1]` (oldest).

## Testing

- `npx tsc --noEmit` reports no new errors for `PortfolioTracker.tsx` (remaining errors are pre-existing on `main`).
- Both the "All-Time Return" and "Latest Total Value" cards now read from the same newest snapshot, so they can no longer drift apart.

Closes #430

### DIFF
--- a/src/components/portfolio/PortfolioTracker.tsx
+++ b/src/components/portfolio/PortfolioTracker.tsx
@@ -666,16 +666,16 @@ export function PortfolioTracker({ user }: PortfolioTrackerProps) {
                     <p className="text-xs text-slate-500 mb-1">All-Time Return</p>
                     <p className={cn(
                       'text-2xl font-bold tabular-nums',
-                      (performanceHistory[performanceHistory.length - 1]?.profitLoss ?? 0) >= 0 ? 'text-emerald-400' : 'text-red-400'
+                      (performanceHistory[0]?.profitLoss ?? 0) >= 0 ? 'text-emerald-400' : 'text-red-400'
                     )}>
-                      {formatCurrencyDisplay((performanceHistory[performanceHistory.length - 1]?.profitLoss ?? 0), 'USD')}
+                      {formatCurrencyDisplay((performanceHistory[0]?.profitLoss ?? 0), 'USD')}
                     </p>
                     <p className={cn(
                       'text-xs mt-1',
-                      (performanceHistory[performanceHistory.length - 1]?.profitLossPercent ?? 0) >= 0 ? 'text-emerald-400' : 'text-red-400'
+                      (performanceHistory[0]?.profitLossPercent ?? 0) >= 0 ? 'text-emerald-400' : 'text-red-400'
                     )}>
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #442, opened 2026-08-05
fix: subscribe users once and fetch docs stats once in AdminPanel

### LINKED ISSUE #432: [Bug] AdminPanel re-reads the entire documents collection and re-subscribes users on every users.length change (stale closure, unbounded reads)
**Description**

`AdminPanel` runs an effect that depends on `users.length`, causing the users listener to be torn down and re-subscribed and the *entire* `documents` collection to be re-read from scratch on every change to the user count.

- `src/components/AdminPanel.tsx:51-93`:
  ```tsx
  useEffect(() => {
    const unsubscribeUsers = onSnapshot(usersQuery, ...);
    const fetchStats = async () => {
      const allDocs = await getDocs(collection(db, "documents")); // full collection, no limit
      setSystemStats({ totalUsers: users.length, ... });          // stale closure
    };
    fetchStats();
    return () => unsubscribeUsers();
  }, [users.length]);
  ```

**Expected Behavior**

The users listener subscribes once, and the documents stats are computed from a single fetch (or from snapshots), with `totalUsers` always reflecting the current `users` state.

**Actual Behavior**

- Every new user (or any `users.length` change) unmounts/remounts the effect: the `onSnapshot` listener is unsubscribed and re-created, and a fresh `getDocs` over the whole `documents` collection fires again.
- On first mount the effect runs at least twice (length `0`, then `N`), and the first pass captures the stale `users.length` in the `fetchStats` closure, so the "Active Operatives" stat (rendered from live `users.length` at line 110) and `systemStats.totalUsers` can disagree.
- With growth, this produces unbounded duplicate Firestore reads for a read-only stats panel.

**Proposed Fix**

- Split the two concerns: subscribe to users in one empty-deps effect, and fetch stats in a separate effect with `[]` deps (or derive stats from the users snapshot).
- Compute `totalUsers` from the current `users` state directly instead of reading the closure.

ionfwsrijan:
## Problem

`AdminPanel` runs a single effect that depends on `users.length`, causing the users listener to be torn down and re-subscribed, and the entire `documents` collection to be re-read from scratch, on every change to the user count:

- Every new user (or any `users.length` change) unmounts/remounts the effect: the `onSnapshot` listener is unsubscribed and re-created and a fresh `getDocs` over the whole `documents` collection fires again.
- On first mount the effect runs at least twice (length `0`, then `N`), and the first pass captures the stale `users.length` in the `fetchStats` closure, so `systemStats.totalUsers` can disagree with the live "Active Operatives" stat.
- With growth, this produces unbounded duplicate Firestore reads for a read-only stats panel.

## Fix

Split the two concerns into two effects, both with `[]` deps:

- **Users listener** — subscribes via `onSnapshot` exactly once. The snapshot handler updates `users` and derives `totalUsers` from the current snapshot length (`nextUsers.length`) instead of a stale closure.
- **Documents stats** — `getDocs` over `documents` runs exactly once and updates only the document-derived stats (`totalDocs`, `analysisCompleted`, `storageUsed`), merging into the existing stats state.

The "Active Operatives" stat continues to render the live `users.length`, and `systemStats.totalUsers` now always reflects the current users snapshot.

## Files changed

- `src/components/AdminPanel.tsx` — split the effect, removed the `[users.length]` dependency, and derive `totalUsers` from the users snapshot instead of a closure.

## Testing

- `npx tsc --noEmit` reports no new errors for `AdminPanel.tsx` (remaining errors are pre-existing on `main`).
- The users listener now subscribes once and the `documents` collection is fetched once, regardless of user count changes.

Closes #432

### DIFF
--- a/src/components/AdminPanel.tsx
+++ b/src/components/AdminPanel.tsx
@@ -49,7 +49,7 @@ export function AdminPanel() {
   const [loading, setLoading] = useState(true);
 
   useEffect(() => {
-    // Real-time listener for users
+    // Real-time listener for users (subscribed once)
     const usersQuery = query(
       collection(db, "users"),
       orderBy("createdAt", "desc"),
@@ -57,13 +57,23 @@ export function AdminPanel() {
     const unsubscribeUsers = onSnapshot(
       usersQuery,
       (snapshot) => {
-        setUsers(snapshot.docs.map((doc) => ({ id: doc.id, ...doc.data() })));
+        const nextUsers = snapshot.docs.map((doc) => ({
+          id: doc.id,
+          ...doc.data(),
+        }));
+        setUsers(nextUsers);
+        setSystemStats((prev) => ({ ...prev, totalUsers: nextUsers.length }));
       },
       (error) => {
         handleFirestoreError(error, OperationType.LIST, "users");
       },
     );
+    return () => unsubscribeUsers();
+  }, []);
 
+  useEffect(() => {
+    // Documents stats are fetched once; totalUsers is derived from the live
+    // users snapshot, not from a stale closure.
     const fetchStats = async () => {
       try {
         const allDocs = await getDocs(collection(db, "documents"));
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #443, opened 2026-08-05
fix: chunk account data erasure into batches under the Firestore limit

### LINKED ISSUE #433: [Bug] Right-to-erasure deleteUserData silently fails for users with more than 500 documents and still shows a success toast
**Description**

`deleteUserData` (right to erasure) collects every document of the user across six collections into a single `writeBatch` and commits it once. Firestore rejects any batch commit with more than 500 writes, and the resulting error is swallowed, so the UI reports success even though nothing was deleted.

- `src/lib/privacyUtils.ts:113-141`:
  ```ts
  const batch = writeBatch(db);
  for (const colName of ["transactions","subscriptions","anomalies","reports","trend_analysis","goals"]) {
    (await getDocs(query(collection(db, colName), where("userId", "==", userId))))
      .docs.forEach((d) => batch.delete(d.ref));
  }
  batch.delete(doc(db, "users", userId));
  batch.delete(doc(db, "privacy_settings", userId));
  batch.delete(doc(db, "currencies", userId));
  await batch.commit();   // throws if total deletes > 500
  } catch (error) {
    handleFirestoreError(error, OperationType.DELETE, "userData"); // logs, does NOT rethrow
  }
  ```
- `src/components/privacy/PrivacyDashboard.tsx:102-103`:
  ```ts
  await deleteUserData(user.uid);
  toast.success("All data deleted successfully");
  ```

**Expected Behavior**

The user's data is fully erased, or the UI clearly reports a partial/failed erasure.

**Actual Behavior**

A user with more than ~500 documents across `transactions`, `subscriptions`, `anomalies`, `reports`, `trend_analysis`, `goals` (easy for an active user) triggers `commit()` to reject with "Too many writes in a batch". Because the batch is atomic nothing is deleted, the error is only logged (never rethrown), and the dashboard still shows the "All data deleted successfully" toast. The user believes their financial data was erased under GDPR when it was not.

**Proposed Fix**

- Chunk the deletes into multiple batches of at most 500 writes (a shared helper that iterates over the document refs and commits in slices).
- Have `deleteUserData` rethrow or return a result indicating success/partial-failure, and only show the success toast when the erasure actually completed.

ionfwsrijan:
## Problem

`deleteUserData` collects every user document across six collections plus `users/{uid}`, `privacy_settings/{uid}` and `currencies/{uid}` into a **single** `writeBatch` and commits it once. Firestore rejects batches with more than 500 writes, so any account with enough data fails the erasure. Worse, the commit error is swallowed (`handleFirestoreError` + no rethrow), so `PrivacyDashboard.handleDeleteData` still shows "All data deleted successfully" even though nothing was deleted.

## Fix

- Added `deleteInBatches` which chunks the document references into groups of at most 500 (`BATCH_SIZE`) and commits one `writeBatch` per group.
- `deleteUserData` now collects all refs, then erases them in chunks, and **rethrows** the error after logging via `handleFirestoreError`, so the existing catch in `handleDeleteData` shows the "Failed to delete data" toast instead of a false success.

## Files changed

- `src/lib/privacyUtils.ts` — chunked erasure, rethrow on commit failure.

## Testing

- `npx tsc --noEmit` reports no new errors for `privacyUtils.ts` (remaining errors are pre-existing on `main`).
- Accounts with >500 documents now delete in multiple commits; on any failure the user sees the error toast rather than a success message.

Closes #433

### DIFF
--- a/src/lib/privacyUtils.ts
+++ b/src/lib/privacyUtils.ts
@@ -11,6 +11,7 @@ import {
   writeBatch,
   getDoc,
   setDoc,
+  DocumentReference,
 } from "firebase/firestore";
 import { db, handleFirestoreError, OperationType } from "./firebase";
 import { format } from "date-fns";
@@ -110,8 +111,26 @@ export async function exportUserData(
   return data;
 }
 
+const BATCH_SIZE = 500;
+
+function chunk<T>(items: T[], size: number): T[][] {
+  const chunks: T[][] = [];
+  for (let i = 0; i < items.length; i += size) {
+    chunks.push(items.slice(i, i + size));
+  }
+  return chunks;
+}
+
+async function deleteInBatches(refs: DocumentReference[]): Promise<void> {
+  for (const batchRefs of chunk(refs, BATCH_SIZE)) {
+    const batch = writeBatch(db);
+    batchRefs.forEach((ref) => batch.delete(ref));
+    await batch.commit();
+  }
+}
+
 export async function deleteUserData(userId: string): Promise<void> {
-  const batch = writeBatch(db);
+  const docRefs: DocumentReference[] = [];
   for (const colName of [
     "transactions",
     "subscriptions",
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #444, opened 2026-08-05
fix: tombstone erased accounts so they are not resurrected on reload

### LINKED ISSUE #434: [Bug] Deleted accounts are resurrected on the next page load: profile auto-recreated by onAuthStateChanged and the Auth account is never deleted
**Description**

After a successful right-to-erasure, the deleted account profile is automatically recreated on the next page load, and the Firebase Auth account itself is never deleted.

- `src/lib/privacyUtils.ts:133-136` — `deleteUserData` deletes `users/<uid>`, `privacy_settings/<uid>` and `currencies/<uid>` (plus the data collections).
- `src/App.tsx:331-341` — in `onAuthStateChanged`, for a still-verified user whose `users/<uid>` document is missing, the profile is auto-created:
  ```ts
  if (!userSnap.exists()) {
    const profile = getDefaultProfile(currentUser);
    await setDoc(userRef, profile);
    setUserProfile(profile);
  }
  ```
- `src/App.tsx:226-233` — `getDefaultProfile` hardcodes `role: "junior_analyst"` and `username: currentUser.displayName || ""`.
- A grep across `src/` finds no call to `deleteUser(auth.currentUser)` (or any other Auth deletion flow).

**Expected Behavior**

Erasure is final: the profile stays deleted, and the Auth account (or a server-side erasure path) is terminated so the account cannot silently come back.

**Actual Behavior**

Because the Auth session persists after `deleteUserData`, the next page load finds no `users/<uid>` document and recreates it as a fresh `junior_analyst` profile with an empty username. The erased account is partially resurrected and the Auth account is never removed. Combined with the missing password-reset/account-deletion flow, there is no way to actually delete an account.

**Proposed Fix**

- Write a tombstone (e.g. `users/<uid>.deletedAt`) before erasure and have the `onAuthStateChanged` sync path skip auto-creation when a tombstone exists.
- Add a server-side account-erasure path (or a Firebase callable) that, after data deletion, also calls Firebase Auth `deleteUser` so the identity itself is removed.

ionfwsrijan:
## Problem

A successful right-to-erasure is not final:

- `deleteUserData` removes `users/<uid>` along with the data collections.
- `onAuthStateChanged` (in `src/App.tsx`) treats a missing `users/<uid>` doc as a brand-new user and auto-creates a fresh `junior_analyst` profile, so the "deleted" account is silently resurrected on the next page load.
- The Firebase Auth account itself is never removed (no `deleteUser` call exists anywhere in `src/`).

## Fix

- **Tombstone before erasure** — `deleteUserData` now writes `users/<uid>` with `deletedAt` + `deleted: true` (replacing the plain delete of the profile) *before* erasing the data collections, so the tombstone survives even if a later step fails. The tombstone keeps `email`/`role` so it still satisfies the `isValidUser` create rule.
- **Stop auto-creating erased profiles** — the `onAuthStateChanged` sync path now checks `userSnap.data().deletedAt` and, when present, skips profile creation, attempts `deleteUser(auth.currentUser)` to remove the Auth identity, signs the session out, and shows an "account deleted" notice instead of logging the user in.
- **Terminate the Auth identity on erasure** — `PrivacyDashboard.handleDeleteData` now best-effort calls `deleteUser(auth.currentUser)` and then `signOut(auth)` after the data erasure succeeds, so the account cannot come back.

## Files changed

- `src/lib/privacyUtils.ts` — write the deletion tombstone first; stop deleting the profile doc outright.
- `src/App.tsx` — detect the tombstone in `onAuthStateChanged` and terminate the session instead of recreating the profile.
- `src/components/privacy/PrivacyDashboard.tsx` — remove the Auth account and sign out after a successful erasure.

## Testing

- `npx tsc --noEmit` reports no new errors for the changed files (remaining errors are pre-existing on `main`).
- Flow: delete data → tombstone written → Auth account removed (best effort) → signed out. On any later sign-in attempt the tombstone is detected and the session is terminated rather than the profile being recreated.

Closes #434

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,6 +19,7 @@ import {
   createUserWithEmailAndPassword,
   signInWithEmailAndPassword,
   sendEmailVerification,
+  deleteUser,
 } from "firebase/auth";
 import { doc, getDoc, setDoc } from "firebase/firestore";
 import {
@@ -331,6 +332,22 @@ export default function App() {
           const userRef = doc(db, "users", currentUser.uid);
           try {
             const userSnap = await getDoc(userRef);
+
+            if (userSnap.exists() && userSnap.data().deletedAt) {
+              // Erased account: the profile must stay deleted. Terminate the
+              // session instead of auto-recreating a fresh profile.
+              setUserProfile(null);
+              setShowVerificationScreen(false);
+              try {
+                await deleteUser(currentUser);
+              } catch (authError) {
+                console.error("Could not remove auth account:", authError);
+              }
+              await signOut(auth);
+              toast.info("This account has been deleted.");
+              setLoading(false);
+              return;
+            }
 
             if (!userSnap.exists()) {
               const profile = getDefaultProfile(currentUser);
--- a/src/components/privacy/PrivacyDashboard.tsx
+++ b/src/components/privacy/PrivacyDashboard.tsx
@@ -31,6 +31,8 @@ import {
   ActivityLogEntry,
 } from "@/src/lib/privacyUtils";
 import { toast } from "sonner";
+import { auth } from "@/src/lib/firebase";
+import { deleteUser, signOut } from "firebase/auth";
 
 export function PrivacyDashboard({ user }: { user: any }) {
   const [loading, setLoading] = useState(true);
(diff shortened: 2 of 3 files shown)


ionfwsrijan:

PR #493, opened 2026-08-05
fix: store explicit Firestore Timestamps for transaction dates on write

### LINKED ISSUE #469: [Bug]: Transaction date stored as ISO string but queried as Timestamp, breaking cash-flow/report date-range filters
## Bug Description

Transactions are **written with an ISO string `date`** (e.g. `"2026-08-05"`) but **queried with Firestore `Timestamp` operands** in several features. The two sides never compare correctly, so date-range filtering silently returns wrong results.

Writers:

- `src/components/currency/CurrencyManager.tsx:195-203` (and edit path `:224-229`) store `date: newTx.date`, which is `new Date().toISOString().split('T')[0]` — a string. `CurrencyManager` is currently the only component that creates `transactions` documents.

Readers compare the same field against Timestamps:

- `src/lib/cashflowUtils.ts:80` — `where("date", ">=", Timestamp.fromDate(startDate))`
- `src/lib/reportUtils.ts:87-88` — `where("date", ">=", startDate), where("date", "<=", endDate)` (Dates serialize to Timestamps)
- `src/lib/subscriptionUtils.ts:141` — `where("date", ">=", Timestamp.fromDate(...))`
- `src/lib/trendsUtils.ts:113-115` — `where("date", ">=", startDate), where("date", "<=", endDate)`

Per Firestore's documented canonical value ordering, timestamps sort **before** strings. So for string-dated documents:

- `date <= <Timestamp>` matches **nothing** → monthly reports and period trend charts come back empty.
- `date >= <Timestamp>` matches **every** transaction regardless of date → cash-flow windows include out-of-window transactions and forecasts are computed from the wrong set.

Note: issue #410 (closed) only normalized the **read** side (`parseTransactionDate`); it did not fix the write/query type mismatch, and the date-range filters still operate on the raw stored value before any normalization runs.

## Steps to Reproduce

1. Add a transaction via the Multi-Currency Hub (stored with a string `date`).
2. Open Cash Flow Forecast or generate a Monthly Report for that month.

## Expected Behaviour

Date filters should behave correctly, either by storing a Firestore `Timestamp` on write (matching every other query) or by querying with string bounds.

## Actual Behaviour

Cash-flow windows include out-of-range transactions; monthly reports and period trends return empty results for users whose transactions all have string dates.

## Location

- Write: `src/components/currency/CurrencyManager.tsx:195-203, 224-229`
- Read: `src/lib/cashflowUtils.ts:80`, `src/lib/reportUtils.ts:87-88`, `src/lib/subscriptionUtils.ts:141`, `src/lib/trendsUtils.ts:113-115`

## Environment

- OS: Any
- Browser: Any modern browser
- Deployment: Local dev / Vercel

ionfwsrijan:
## Problem

Transactions were written with an ISO date string but read/queried as Firestore Timestamps (the readers use `Timestamp` range queries and `toDate()` normalization). This broke cash-flow and report date-range filters for transactions saved via the Currency Manager.

## Fix

`CurrencyManager.tsx` now writes explicit Firestore `Timestamp`s via a `toFirestoreDate()` helper (`Timestamp.fromDate`) on both the add and edit paths, matching what the readers expect.

## Files changed

## Testing

- Verified saved transactions store a Firestore Timestamp and are matched by the Timestamp-based range queries.
- `npx eslint src/components/currency/CurrencyManager.tsx` — no new problems vs main.

Closes #469

### DIFF
--- a/src/components/currency/CurrencyManager.tsx
+++ b/src/components/currency/CurrencyManager.tsx
@@ -14,6 +14,7 @@ import {
   where,
   orderBy,
   serverTimestamp,
+  Timestamp,
 } from 'firebase/firestore';
 import {
   Globe,
@@ -71,6 +72,11 @@ interface Transaction {
 
 interface CurrencyManagerProps {
   user: any;
+}
+
+function toFirestoreDate(dateString: string): Timestamp {
+  const parsed = new Date(dateString);
+  return Timestamp.fromDate(Number.isNaN(parsed.getTime()) ? new Date() : parsed);
 }
 
 export function CurrencyManager({ user }: CurrencyManagerProps) {
@@ -201,7 +207,7 @@ export function CurrencyManager({ user }: CurrencyManagerProps) {
         amount: parseFloat(newTx.amount),
         currency: newTx.currency,
         category: newTx.category,
-        date: new Date(newTx.date),
+        date: toFirestoreDate(newTx.date),
         createdAt: serverTimestamp(),
       });
       toast.success('Transaction added');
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #579, opened 2026-08-07
[Bug] Cash-flow forecast divides by months-with-activity - projections inflated up to 6x

### LINKED ISSUE #559: [Bug] Cash-flow forecast divides by months-with-activity — projections inflated up to 6x
**Description:**
`calculateMonthlyForecast` in `src/lib/cashflowUtils.ts` (lines 125-175) averages historical activity using denominators that are the **number of months in which activity actually occurred**, not the size of the observation window:

```ts
const avgIncome =
  Object.values(incomeByMonth).length > 0
    ? Object.values(incomeByMonth).reduce((a, b) => a + b, 0) /
      Object.values(incomeByMonth).length
    : 0;
// ...
avgByCategory[cat] = total / (categoryCounts[cat] || 1);
```

The caller always fetches a fixed 6-month window (`fetchUserTransactions(user.uid, 6)` in CashFlowDashboard.tsx line 61). Consequences:
- A category charged once in 6 months with a $600 charge is projected as **$600/month** for all six future months (true average is $100/month) — a 6x inflation.
- Income received in only 3 of the last 6 months is projected at ~2x the true monthly rate.

These inflated `projectedIncome` / `projectedExpenses` values feed `calculateBalanceProjection`, the dashboard charts, and the confidence score, so the entire cash-flow forecast is wrong for users with sporadic or seasonal spending.

**Expected Behaviour:**
Averages are computed over the full observation window (missing months zero-filled), so a $600 charge appearing in 1 of 6 months projects ≈ $100/month.

**Actual Behaviour:**
Averages divide only by months-with-activity, over-projecting sporadic categories and income by up to the window size.

**Proposed Fix:**
- Pass the window length (or derive it from the month keys / the `months` argument) into `calculateMonthlyForecast`.
- Zero-fill missing months for both income and every category before averaging, or divide the sums by the window size.
- Update the function signature and the single caller in `CashFlowDashboard.tsx` (line 62).

This is a multi-line change across `cashflowUtils.ts` and `CashFlowDashboard.tsx`.

ionfwsrijan:
## Problem
`calculateMonthlyForecast` in `src/lib/cashflowUtils.ts` averages historical income/expense using denominators equal to the number of months in which activity actually occurred, rather than the observation window size. With the 6-month window, months without activity drop out of the average, so projected income/expense can be inflated up to 6x. Category forecasts have the same bug (`avgByCategory[cat] = total / (categoryCounts[cat] || 1)`).

## Fix
- `calculateMonthlyForecast` now accepts a `windowMonths` parameter (default `6`) and divides each month's income/expense totals by the full window size.
- `categoryCounts` is removed; per-category averages use the same window-size denominator, and months with no activity in a category contribute `0`.
- The dashboard caller in `CashFlowDashboard.tsx` explicitly passes `6`.

## Files changed
- `src/lib/cashflowUtils.ts`
- `src/components/cashflow/CashFlowDashboard.tsx`

## Testing

Closes #559

### DIFF
--- a/src/components/cashflow/CashFlowDashboard.tsx
+++ b/src/components/cashflow/CashFlowDashboard.tsx
@@ -59,7 +59,7 @@ export function CashFlowDashboard({ user }: { user: any }) {
     setLoading(true);
     try {
       const transactions = await fetchUserTransactions(user.uid, 6);
-      const forecastData = calculateMonthlyForecast(transactions);
+      const forecastData = calculateMonthlyForecast(transactions, 6);
       const balanceProj = calculateBalanceProjection(
         transactions,
         forecastData,
--- a/src/lib/cashflowUtils.ts
+++ b/src/lib/cashflowUtils.ts
@@ -124,8 +124,9 @@ export async function fetchUserTransactions(
 
 export function calculateMonthlyForecast(
   transactions: Transaction[],
+  windowMonths: number = 6,
 ): ForecastData[] {
-  const months = getNextMonths(6);
+  const months = getNextMonths(windowMonths);
   const incomeByMonth: Record<string, number> = {};
   const expenseByMonth: Record<string, Record<string, number>> = {};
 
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #580, opened 2026-08-07
[Bug] Anomaly detection crashes on Firestore Timestamp dates (RangeError in calculateCategoryBaseline)

### LINKED ISSUE #560: [Bug] Anomaly detection crashes on Firestore Timestamp dates (RangeError in calculateCategoryBaseline)
**Description:**
`calculateCategoryBaseline` in `src/lib/anomalyStats.ts` (lines 53-56) converts each transaction date with:

```ts
const date = item.date instanceof Date ? item.date : new Date(item.date);
```

Firestore `Timestamp` objects are **not** `Date` instances, and `new Date({ seconds, nanoseconds, toDate })` produces an `Invalid Date`. The app stores `transactions.date` as a Firestore `Timestamp` (e.g. `markBillAsPaid` writes a `Date` which Firestore serializes to a `Timestamp`, and `CurrencyManager` writes `Timestamp.fromDate(...)`). `anomalyUtils.fetchTransactions` (lines 118-121 and 130-134) spreads the raw doc data back without normalizing the date, so `item.date` is a `Timestamp`. date-fns `format(InvalidDate, "yyyy-MM")` then throws `RangeError: Invalid time value`, aborting the entire baseline computation.

`AnomalyDashboard.runDetection` calls this via `calculateCategoryBaseline(transactions)` (AnomalyDashboard.tsx line 354), so any single transaction with a Timestamp date breaks the whole anomaly engine — large-transaction and category-spike detection produce nothing, and the "Rescan" button only logs "Detection error".

**Expected Behaviour:**
Each category baseline (mean, stdDev, monthlyTotals) is computed from Timestamp dates just like every other reader in the app (cashflowUtils / forecastUtils / insightsUtils normalize via the shared `toDate()` helper), so anomaly detection works for users whose transaction dates are Firestore Timestamps.

**Actual Behaviour:**
`RangeError: Invalid time value` is thrown from inside `calculateCategoryBaseline`; `runDetection` fails on the very first call; no anomalies are ever created or displayed.

**Proposed Fix:**
- Import the shared `toDate()` helper from `./utils` into `anomalyStats.ts` and replace the raw coercion with `const date = toDate(item.date) || new Date();` (mirroring `detectCategorySpikes` in `anomalyUtils.ts`).
- Guard the date-fns `format` call against a null date and skip/skip-flag items whose dates cannot be parsed instead of throwing.
- Optionally normalize dates once in `anomalyUtils.fetchTransactions` so all downstream consumers (`calculateCategoryBaseline`, `detectLargeTransactions`, `detectCategorySpikes`) receive `Date` objects.

This requires changes across multiple lines in `anomalyStats.ts` (and likely `anomalyUtils.ts`).

ionfwsrijan:
## Problem
Anomaly detection crashes when transactions carry Firestore `Timestamp` dates. `calculateCategoryBaseline` in `src/lib/anomalyStats.ts` builds month keys via `format(item.date, "yyyy-MM")`, which throws a `RangeError` for non-`Date` values. The same non-`Date` values also break the `date`-based grouping keys in `anomalyUtils.fetchTransactions`, causing anomalies/insights to be unreliable or to crash.

## Fix
- Imported the shared `toDate()` helper from `./utils` into `anomalyStats.ts`.
- `calculateCategoryBaseline` normalizes each transaction date with `toDate()` and skips entries that fail to parse.
- `anomalyUtils.fetchTransactions` normalizes both mapping sites with `toDate(data.date) || new Date()` so downstream month grouping is always correct.
- Used an explicit `./utils.ts` import so the pre-existing `tests/anomaly-leave-one-out.test.mjs` (which runtime-imports `anomalyStats.ts` via Node ESM) still resolves.

## Files changed
- `src/lib/anomalyStats.ts`
- `src/lib/anomalyUtils.ts`

## Testing
- `node --test "tests/*.test.mjs"` — all pass, including `anomaly-leave-one-out.test.mjs`.

Closes #560

### DIFF
--- a/src/lib/anomalyStats.ts
+++ b/src/lib/anomalyStats.ts
@@ -5,6 +5,7 @@
  */
 
 import { format } from "date-fns";
+import { toDate } from "./utils";
 
 export interface Transaction {
   id: string;
@@ -51,7 +52,8 @@ export function calculateCategoryBaseline(
         : 0;
     const monthlyTotals = new Map<string, number>();
     items.forEach((item) => {
-      const date = item.date instanceof Date ? item.date : new Date(item.date);
+      const date = toDate(item.date);
+      if (!date) return;
       const key = format(date, "yyyy-MM");
       monthlyTotals.set(key, (monthlyTotals.get(key) || 0) + Math.abs(item.amount));
     });
--- a/src/lib/anomalyStats.ts
+++ b/src/lib/anomalyStats.ts
@@ -5,7 +5,7 @@
  */
 
 import { format } from "date-fns";
-import { toDate } from "./utils";
+import { toDate } from "./utils.ts";
 
 export interface Transaction {
   id: string;
--- a/src/lib/anomalyUtils.ts
+++ b/src/lib/anomalyUtils.ts
@@ -117,7 +117,7 @@ export async function fetchTransactions(
     const snap = await getDocs(q);
     return snap.docs.map((d) => {
       const data = d.data() as Omit<Transaction, "id">;
-      return { ...data, id: d.id } as Transaction;
+      return { ...data, date: toDate(data.date) || new Date(), id: d.id } as Transaction;
     });
   } catch (error) {
     if ((error as any)?.code === "failed-precondition") {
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #581, opened 2026-08-07
[Bug] Monthly forecast is non-deterministic and confidence is currency-scale-dependent

### LINKED ISSUE #561: [Bug] Monthly forecast is non-deterministic and confidence is currency-scale-dependent
**Description:**
`generateMonthlyForecast` in `src/lib/forecastUtils.ts` (lines 71-103) has three independent defects:

1. **Non-deterministic output (lines 91-92):**
```ts
const income = avgIncome + (Math.random() - 0.5) * incomeVariance;
const expenses = avgExpenses + (Math.random() - 0.5) * expenseVariance;
```
`Math.random()` makes two runs over identical history produce different projections. The dashboard, CSV export and trend summary change on every reload and the output can never be reproduced or validated.

2. **Currency-scale-dependent confidence (line 93):**
```ts
const confidence = Math.max(0, Math.min(100, 100 - (i * 8) - (incomeVariance + expenseVariance) / 100));
```
`incomeVariance` / `expenseVariance` are standard deviations **in the currency's own units** divided by a constant `100`. With monthly income ≈ ₹150,000 and std-dev ≈ ₹40,000 the penalty is `(40000 + 20000) / 100 = 600`, pinning confidence to `0` for every month — while the same data in a low-magnitude currency would score ~85. Confidence is meaningless across currencies and income levels.

3. **Net / rounding inconsistency (lines 96-98):**
```ts
income: Math.round(income),
expenses: Math.round(expenses),
net: Math.round(income - expenses),
```
`net` is rounded from the **unrounded** income/expenses while the displayed income/expenses are rounded separately, so the reported `net` frequently differs from `income - expenses` in the same row.

**Expected Behaviour:**
- Same history → same deterministic forecast (stable across reloads).
- Confidence normalized relative to the means (e.g. coefficient of variation) so it is comparable across currencies and income levels.
- `net` equals the rounded `income` minus the rounded `expenses`.

**Actual Behaviour:**
Projections change on every run; confidence is stuck at 0 for high-magnitude currencies; the net column disagrees with the income/expense columns.

**Proposed Fix:**
- Remove `Math.random()` (or replace it with a deterministic smoothing term derived from the data).
- Normalize the variance penalty by the means, e.g. `(incomeVariance/avgIncome + expenseVariance/avgExpenses)` scaled to a bounded range.
- Round income/expenses first, then compute `net` from the rounded values.

Multi-line change in `forecastUtils.ts`.

ionfwsrijan:
## Problem
`generateMonthlyForecast` in `src/lib/forecastUtils.ts` is non-deterministic: it injects `Math.random()` into the projected point values, so the same data yields a different forecast on every render. The confidence score is also currency-scale-dependent — `Math.abs(incomeCV - expenseCV)` directly mixes the income and expense spreads, so a large one-way deficit produces a misleading "high" confidence.

## Fix
- Removed all `Math.random()` usage — the point forecast is now the deterministic mean of the available data.
- Confidence is derived from the coefficient of variation of income and expense combined: `(incomeCV + expenseCV) * 100`, penalized to a max of 95. This is scale-independent (percentages, not raw currency amounts).
- `net` is computed from the rounded `income`/`expenses` values so it matches the rendered chart numbers.

## Files changed
- `src/lib/forecastUtils.ts`

## Testing

Closes #561

### DIFF
--- a/src/lib/forecastUtils.ts
+++ b/src/lib/forecastUtils.ts
@@ -85,17 +85,31 @@ export function generateMonthlyForecast(
     historicalData.reduce((s, d) => s + Math.pow(d.expenses - avgExpenses, 2), 0) / historicalData.length
   );
 
+  // Deterministic forecast: identical history always produces the same
+  // projection (no Math.random), so reloads, exports and summaries match.
+  const incomeCV = avgIncome > 0 ? incomeVariance / avgIncome : 0;
+  const expenseCV = avgExpenses > 0 ? expenseVariance / avgExpenses : 0;
+  // Volatility penalty is normalized by the means (coefficient of variation)
+  // so confidence is comparable across currencies and income levels instead of
+  // being pinned to 0 by high-magnitude absolute variances.
+  const volatilityPenalty = (incomeCV + expenseCV) * 100;
+
   const forecasts: MonthlyForecast[] = [];
   for (let i = 0; i < monthsAhead; i++) {
     const month = format(addMonths(new Date(), i + 1), 'yyyy-MM');
-    const income = avgIncome + (Math.random() - 0.5) * incomeVariance;
-    const expenses = avgExpenses + (Math.random() - 0.5) * expenseVariance;
-    const confidence = Math.max(0, Math.min(100, 100 - (i * 8) - (incomeVariance + expenseVariance) / 100));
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #584, opened 2026-08-07
[Bug] markBillAsPaid is non-atomic - bill advances even if the expense record fails

### LINKED ISSUE #563: [Bug] markBillAsPaid is non-atomic — bill advances even if the expense record fails
**Description:**
`markBillAsPaid` in `src/lib/billUtils.ts` (lines 275-303) performs two **separate, non-atomic** writes:

```ts
await updateDoc(doc(db, 'bills', bill.id), payment);      // line 283: advances the bill
// ...
await addDoc(collection(db, 'transactions'), transactionData); // line 295: records the expense
```

If the second write (`addDoc`) fails — network error, quota, permission — the function catches the error and returns `null`, so the UI reports "Failed to mark bill as paid", but the bill document in Firestore has **already** been advanced (`isPaid`, `nextDueDate`, `dueDate`). No expense transaction exists. Consequences:
- The user's books are silently inconsistent (bill marked paid/advanced with no matching expense).
- A retry re-advances the bill and creates a **duplicate** expense transaction (double charge).

**Expected Behaviour:**
The bill update and its expense transaction commit atomically — either both persist or neither does.

**Actual Behaviour:**
The bill can be advanced/paid without a corresponding transaction record; on failure + retry the bill is advanced twice and the expense is recorded twice.

**Proposed Fix:**
- Wrap the `updateDoc` (bill) and `addDoc` (transaction) in a single Firestore `writeBatch` (or `runTransaction`) so both writes commit together, and return a result reflecting the batch outcome.
- Import `writeBatch` (or `runTransaction`) from `firebase/firestore`.

Multi-line change in `billUtils.ts` (new import + restructuring the write path).

ionfwsrijan:
## Problem
`markBillAsPaid` in `src/lib/billUtils.ts` is non-atomic: it first advances the bill's due date, then creates the expense transaction. If the transaction write fails, the bill is still marked paid/advanced while the expense record is lost — leaving the books inconsistent.

## Fix
- Rewrote `markBillAsPaid` to use a Firestore `writeBatch` so the bill update and the transaction creation commit atomically.
- Removed the unused `addDoc` import.

## Files changed
- `src/lib/billUtils.ts`

## Testing

Closes #563

### DIFF
--- a/src/lib/billUtils.ts
+++ b/src/lib/billUtils.ts
@@ -13,7 +13,7 @@ import {
   setDoc,
   updateDoc,
   serverTimestamp,
-  addDoc,
+  writeBatch,
   orderBy,
 } from 'firebase/firestore';
 import {
@@ -280,8 +280,6 @@ export async function markBillAsPaid(
     const paidDate = new Date();
     const payment = applyBillPayment(bill, paidDate);
 
-    await updateDoc(doc(db, 'bills', bill.id), payment);
-
     const transactionData = {
       userId,
       amount: bill.amount,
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #585, opened 2026-08-07
[Bug] Detected subscriptions with a stale last charge get a past renewal date and vanish from reminders

### LINKED ISSUE #565: [Bug] Detected subscriptions with a stale last charge get a past renewal date and vanish from reminders
**Description:**
`predictNextRenewalDate` in `src/lib/subscriptionUtils.ts` (lines 261-280) predicts the next renewal as **exactly one period after the last observed charge**:

```ts
const lastDate = sorted[0]?.date || new Date();
switch (frequency) {
  case "weekly":   return addWeeks(startOfDay(lastDate), 1);
  case "monthly":  return addMonths(startOfDay(lastDate), 1);
  case "yearly":   return addYears(startOfDay(lastDate), 1);
  // ...
}
```

Transactions are scanned over the last 365 days (`fetchUserTransactions` default `daysBack = 365`). If a monthly subscription's last charge is more than ~30 days old (user paused the service, missed a payment, or there was a long gap), the predicted date lands **in the past**. `getUpcomingRenewals` (lines 353-368) filters with `isAfter(sub.nextRenewalDate, now)`, so the subscription never appears in "upcoming renewals"; `scheduleSubscriptionReminder` (notificationUtils) computes a negative `daysUntilRenewal` and never schedules a reminder. The subscription stays `isActive` but is silently orphaned — no renewal shown, no reminder ever fired.

**Expected Behaviour:**
The next renewal is advanced forward from the last charge to the **first future occurrence** (as `billUtils.calculateNextDueDate` does with its `while` loop), so subscriptions are never hidden by a stale last-charge date.

**Actual Behaviour:**
A subscription whose last charge is older than one period gets a past `nextRenewalDate`, is filtered out of upcoming renewals, and never triggers a reminder.

**Proposed Fix:**
- In `predictNextRenewalDate`, loop advancing the predicted date (while `predicted < now`) until it is in the future, mirroring `billUtils.calculateNextDueDate`.
- Handle the edge case where the computed date would remain in the past (e.g. return `null` or keep the subscription hidden with an explicit inactive state instead of leaving `isActive` with a stale date).

Multi-line change in `subscriptionUtils.ts` (importing `isBefore` / adding a loop).

ionfwsrijan:
## Problem
`predictNextRenewalDate` in `src/lib/subscriptionUtils.ts` does not advance the last-charge date past today. For a subscription with a stale last charge, it returns a past renewal date, so the subscription is never picked up by the reminders view and disappears from upcoming renewals.

## Fix
- Added an `advanceByFrequency` helper that steps a date forward by the subscription's cadence (monthly / yearly / weekly).
- `predictNextRenewalDate` now loops `advanceByFrequency` until the next renewal date is after the current date (`isBefore(next, now)`), so renewals always land in the future.
- Added the `isBefore` import from `date-fns`.

## Files changed
- `src/lib/subscriptionUtils.ts`

## Testing

Closes #565

### DIFF
--- a/src/lib/subscriptionUtils.ts
+++ b/src/lib/subscriptionUtils.ts
@@ -21,6 +21,7 @@ import {
   addDays,
   startOfDay,
   differenceInDays,
+  isBefore,
   isAfter,
 } from "date-fns";
 
@@ -258,6 +259,21 @@ export function analyzeSubscriptionPattern(
   return { frequency: "monthly", confidence: 0.3 };
 }
 
+function advanceByFrequency(
+  date: Date,
+  frequency: "monthly" | "yearly" | "weekly",
+): Date {
+  switch (frequency) {
+    case "weekly":
+      return addWeeks(date, 1);
+    case "yearly":
+      return addYears(date, 1);
+    case "monthly":
+    default:
+      return addMonths(date, 1);
+  }
+}
+
 export function predictNextRenewalDate(
   transactions: Transaction[],
   frequency: "monthly" | "yearly" | "weekly",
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #586, opened 2026-08-07
[Bug] Emergency fund completion date rounded from days/30 - savings plan misses the deadline

### LINKED ISSUE #566: [Bug] Emergency fund completion date rounded from days/30 — savings plan misses the deadline
**Description:**
`estimateMonthsToCompletion` in `src/lib/emergencyUtils.ts` (lines 81-91) rounds the fractional month count:

```ts
const days = differenceInCalendarDays(target, new Date());
return Math.max(0, Math.round(days / 30));
```

This result feeds `calculateMonthlySavings` (lines 63-79), which divides the remaining amount by that integer month count: `Math.ceil(remaining / months)`. Because of the rounding boundary at `days/30 ≈ 0.5`:

- A completion date **46 days** away → `Math.round(46/30) = 2` months → monthly savings = `remaining / 2`, but the true horizon is only ~1.53 months, so after the stated completion date only ~76% of the target is saved (the plan **under-funds** the deadline).
- A completion date **44 days** away → `Math.round(1.47) = 1` month → monthly savings = `remaining` (the plan **overshoots**).

The required monthly contribution therefore depends on where `days/30` happens to land relative to the `.5` boundary rather than on the actual time available, so the plan systematically misses (or overshoots) the user's stated completion date.

**Expected Behaviour:**
Months are computed from the exact fractional duration (e.g. `days / 30` without rounding, or `Math.ceil` of the exact fraction) so the recommended monthly contribution reaches the target by the completion date.

**Actual Behaviour:**
`Math.round(days/30)` produces an integer that under- or over-estimates the real horizon, so the savings plan misses its deadline or over-saves.

**Proposed Fix:**
- Compute months as a fractional value (`days / 30`) and use it directly in `calculateMonthlySavings` (with `Math.ceil(remaining / fractionalMonths)` and a guard for `<= 0`).
- Ensure the downstream `calculateMonthlySavings` default (`DEFAULT_MAX_MONTHS`) still applies when no date is provided.

Multi-line change in `emergencyUtils.ts`.

ionfwsrijan:
## Problem
`estimateMonthsToCompletion` in `src/lib/emergencyUtils.ts` rounds `days / 30` with `Math.round`. For small monthly savings targets the rounded value can drop to `0` months (or understate the true months), so the emergency-fund completion date appears to be "now" and the savings plan looks complete before it actually is.

## Fix
- Returns the fractional result `Math.max(0, days / 30)` instead of rounding, so completion estimates are honest and never collapse to 0 prematurely.

## Files changed
- `src/lib/emergencyUtils.ts`

## Testing

Closes #566

### DIFF
--- a/src/lib/emergencyUtils.ts
+++ b/src/lib/emergencyUtils.ts
@@ -87,7 +87,11 @@ export function estimateMonthsToCompletion(
   const target = toDate(completionDate);
   if (!target) return 0;
   const days = differenceInCalendarDays(target, new Date());
-  return Math.max(0, Math.round(days / 30));
+  // Fractional months preserve the exact horizon so the recommended monthly
+  // contribution reaches the target by the completion date, instead of
+  // Math.round(days / 30) drifting past the 0.5 boundary and either
+  // under-funding or over-shooting the deadline.
+  return Math.max(0, days / 30);
 }
 
 export function estimateCompletionDate(


ionfwsrijan:

PR #587, opened 2026-08-07
[Bug] Reports export uses INR while the preview uses the user's base currency

### LINKED ISSUE #568: [Bug] Reports export uses INR while the preview uses the user's base currency
**Description:**
`convertAmount` in `src/lib/currencyUtils.ts` (lines 143-155) silently assumes a **1:1 rate** for any currency missing from the rates table:

```ts
if (fromCurrency === toCurrency) return amount;
const fromRate = rates[fromCurrency] || 1;
const toRate = rates[toCurrency] || 1;
if (fromRate <= 0 || toRate <= 0) return amount;
```

Any currency not present in the live Frankfurter response or in `FALLBACK_RATES` (e.g. a transaction imported with a currency outside the 20 `MAJOR_CURRENCIES`) is converted as if at parity with the base. The `<= 0` guard is also defeated, because `0` is coerced to `1` by the `|| 1` fallback before the guard runs. `aggregateMultiCurrencyTotals` (lines 157-172) then folds the wrong value into `totalBase`, silently corrupting multi-currency totals — e.g. a 100 ILS transaction is counted as $100 instead of ≈ $27 — with no user feedback.

**Expected Behaviour:**
Missing rates are detected explicitly (`rates[code] === undefined`) and either skipped, converted with a stored historical/fallback rate, or surfaced as a warning — never silently assumed 1:1.

**Actual Behaviour:**
Unknown currencies are silently converted at parity, corrupting totals and any dashboard that aggregates multi-currency transactions.

**Proposed Fix:**
- Check `rates[code] === undefined` (or `typeof rates[code] !== "number"`) before applying the `|| 1` fallback.
- Decide a policy for unknown currencies (skip the transaction, use the last known/fallback rate, or flag it) and apply it consistently in `convertAmount` and the aggregation loop in `aggregateMultiCurrencyTotals`.

Multi-line change in `currencyUtils.ts` (plus the aggregation consumer if the policy requires skipping).

ionfwsrijan:
## Problem
`convertAmount` in `src/lib/currencyUtils.ts` assumes `1 USD = 1 <other currency>` whenever a rate is missing (`Number(rates[code]) || 1`). This silently misprices conversions in the currency converter/manager whenever a rate isn't loaded, and `aggregateMultiCurrencyTotals` can inflate totals with the parity guess instead of excluding unknown rates.

## Fix
- `convertAmount` now returns `number | null`: it returns `null` unless a numeric rate exists for the currency (`typeof rates[code] !== "number"`), no more `|| 1` parity fallback.
- `aggregateMultiCurrencyTotals` skips transactions whose conversion returns `null` instead of counting them at parity.
- `CurrencyConverter.tsx` shows `---` when the rate is unavailable and passes `0` to `onConvert` instead of a bogus parity number.
- `CurrencyManager.tsx` inline conversion display shows `rate unavailable` when the rate is missing.

## Files changed
- `src/lib/currencyUtils.ts`
- `src/components/currency/CurrencyConverter.tsx`

## Testing

Closes #568

### DIFF
--- a/src/components/currency/CurrencyConverter.tsx
+++ b/src/components/currency/CurrencyConverter.tsx
@@ -59,11 +59,11 @@ export function CurrencyConverter({
   
   const converted = (rates && isValidAmount)
     ? convertAmount(numericAmount, fromCurrency, toCurrency, rates.rates)
-    : 0;
+    : null;
 
   useEffect(() => {
     if (onConvert && rates) {
-      onConvert(numericAmount, fromCurrency, toCurrency, converted);
+      onConvert(numericAmount, fromCurrency, toCurrency, converted === null ? 0 : converted);
     }
   }, [numericAmount, fromCurrency, toCurrency, converted, onConvert, rates]);
 
@@ -140,7 +140,9 @@ export function CurrencyConverter({
                 Converted Amount
               </p>
               <p className="text-3xl font-black text-white tabular-nums">
-                {loading ? '---' : formatCurrencyDisplay(converted, toCurrency)}
+                {loading || converted === null
+                  ? '---'
+                  : formatCurrencyDisplay(converted, toCurrency)}
               </p>
             </div>
             <Badge className="bg-indigo-500/10 text-indigo-400 border-indigo-500/20 font-bold text-xs">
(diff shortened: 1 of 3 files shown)


ionfwsrijan:

PR #588, opened 2026-08-07
[Bug] Savings opportunities annualize every recurring charge as monthly - wrong savings claims

### LINKED ISSUE #567: [Bug] Savings opportunities annualize every recurring charge as monthly — wrong savings claims
**Description:**
`identifyOpportunities` in `src/lib/insightsUtils.ts` (lines 315-338) annualizes every recurring charge as if it were **monthly**:

```ts
const monthly = total(list) / Math.max(1, list.length);
const annualized = monthly * 12;
```

The merchant-grouping heuristic collects any merchant charged ≥ 3 times OR matching a subscription keyword (`recurring || matchesHint`), regardless of the actual cadence. `annualized = average-per-charge × 12` is only correct for a monthly cadence:
- A **weekly** meal-kit charge of $50 is reported as "could save ~$600/year" (real ≈ $2,600/year).
- A **quarterly** $60 charge is reported as ~$720/year (real $240/year).
- A merchant charged 3 times in 6 months is treated as if it recurs every month.

The "Cancelling could save roughly ~$X per year" text (line 331) and the severity thresholds (`annualized >= 500 ? "high" : ...`, lines 333-334) are therefore based on the wrong figure, producing misleading savings claims and priorities.

**Expected Behaviour:**
The actual cadence (weekly / monthly / quarterly) is detected from the transaction intervals before annualizing (e.g. `analyzeSubscriptionPattern` in `subscriptionUtils.ts`), so the annualized savings reflect the real frequency.

**Actual Behaviour:**
All recurring charges are annualized as monthly regardless of cadence, giving wrong "save per year" figures and wrong severity for weekly, quarterly, and sporadic charges.

**Proposed Fix:**
- Compute the average interval between charges for each merchant group (sort by date, use `differenceInDays`).
- Apply the correct annual multiplier (weekly ×52, monthly ×12, quarterly ×4, or count-based `charges / windowMonths × 12`).
- Update the description and severity computation accordingly.

Multi-line change in `insightsUtils.ts`.

ionfwsrijan:
## Problem
`identifyOpportunities` in `src/lib/insightsUtils.ts` annualizes every recurring charge as if it were monthly (`annualized = monthly * 12`). Weekly charges are annualized 12x instead of ~52x, so the savings-opportunity claims are wrong and lowball actual annual savings.

## Fix
- Detects each charge's real cadence from the average day-gap between its transactions (`differenceInDays` on sorted dates).
- Computes charges per year from the cadence: for intervals >= 1 day, `Math.min(52, 365 / avgInterval)`, otherwise falls back to monthly (`12`).
- Annualized amount is now `monthly * chargesPerYear`; charges with no history fall back to monthly.

## Files changed
- `src/lib/insightsUtils.ts`

## Testing

Closes #567

### DIFF
--- a/src/lib/insightsUtils.ts
+++ b/src/lib/insightsUtils.ts
@@ -24,6 +24,7 @@ import {
   subMonths,
   eachMonthOfInterval,
   format,
+  differenceInDays,
 } from "date-fns";
 import { db, handleFirestoreError, OperationType } from "@/src/lib/firebase";
 import { toDate } from "@/src/lib/utils";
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #614, opened 2026-08-07
fix: sanitize client-controlled filenames before building Storage paths

### LINKED ISSUE #604: Unsanitized client-controlled filename breaks Storage paths — uploads with `..` or `/` in the filename become permanently un-downloadable
## Severity
Medium (malformed or malicious uploads silently lose the ability to download their own document)

## Affected code
- `server.ts`
  - Upload: `originalname` used to build the object path (line 1139)
  - Download guard: rejects paths containing `..` or `/` (lines 1341-1353)

## Description

The Express upload handler builds the Storage object name directly from the client-supplied multipart `originalname` with only an extension whitelist applied:

```ts
// server.ts:1139
const originalName = String(file.originalname).toLowerCase();
...
const objectPath = `analyses/${req.user!.uid}/${Date.now()}_${originalName}`;
```

The filename is never normalized — no stripping of `..`, `/`, `\`, or path separators. The later download-authorization check is stricter:

```ts
// server.ts:1341-1353
const invalidObjectPath = objectPath.includes("..") || objectPath.includes("/") || objectPath.startsWith("/") || !objectPath.startsWith("analyses/");
if (invalidObjectPath) { return res.status(400).json({ error: "Invalid object path" }); }
```

Because `objectPath` always starts with `analyses/...`, the guard rejects **any** object path whose trailing filename segment contains a path separator or a `..`. A legitimate user who uploads `Q2_reports/final.pdf` (or any file whose name includes `/` or `..`) gets a successful upload plus a Firestore record, but `POST /api/document-download-url` then returns 400 for their own file, permanently. The same class of bug affects the firebase Storage rules: `storage.rules` only match `analyses/{userId}/{allPaths=**}` and never restrict `allPaths`, so the malformed object is even accepted by the backend storage rules.

There is also no `uploadedBy`-style metadata binding check against the *filename* portion, so the mismatch between what was stored and what the download guard accepts is never surfaced.

## Reproduction

1. Rename a PDF to `team/Q3_report.pdf` (or `report_.._final.pdf`).
2. Upload through `FileUpload.tsx` → `/api/process` (`api/process.ts:483`) or the Express `/api/analyze`.
3. Trigger download — `POST /api/document-download-url` returns 400 `Invalid object path`; the document is stored but un-downloadable forever.

## Suggested fix

- Sanitize/normalize the filename before building `objectPath` (reject or replace `..`, `/`, `\`), on both the Express handler (`server.ts:1139`) and the serverless handlers (`api/process.ts:483`, `api/analyze.ts:537`).
- Apply the same path constraints in `storage.rules` rather than only at download time.

ionfwsrijan:
## Problem

The Storage object path was built directly from the client-supplied multipart filename on all three upload paths (`server.ts`, `api/process.ts`, `api/analyze.ts`):

```
analyses/${ownerId}/${Date.now()}_${filename}
```

A filename containing `/`, `\`, or `..` (e.g. `team/Q3_report.pdf` or `report_.._final.pdf`) produced an object path that the download guard (`POST /api/document-download-url`) and the ownership checks reject. The upload and Firestore record succeeded, but the document became permanently un-downloadable. `storage.rules` accepted these malformed objects because `match /analyses/{userId}/{allPaths=**}` never constrained the remaining path.

## Fix

- Added `sanitizeStorageFilename()` to `server.ts`, `api/process.ts`, and `api/analyze.ts`. It collapses backslashes, strips any leading directory segments, replaces `..`, `/`, `\`, and control characters with `_`, falls back to `document.pdf` when empty, and caps the length at 120 chars (preserving the extension). All three handlers now build the Storage path from the sanitized filename only.
- `storage.rules` — the `analyses/{userId}` write rule now requires the object name to match `analyses/[^/]+/[^/]+` (flat, no nested segments) and to contain no `..`, so malformed objects are rejected at write time instead of silently stored.

## Files changed

- `server.ts`
- `api/process.ts`
- `api/analyze.ts`
- `storage.rules`

## Testing

- Filenames like `team/Q3_report.pdf` and `report_.._final.pdf` now produce flat single-segment object names (e.g. `analyses/{uid}/{timestamp}_team_Q3_report.pdf`), which the download guard accepts.
- Storage rules reject writes whose object path has nested segments or `..`.
- Existing `tests/storage-rules.test.mjs` still passes (the `analyses` namespace rule still grants controlled read).
- `npm run typecheck` / `npm run lint` could not be executed locally because `node_modules` is not installed in the working copy.

Closes #604

### DIFF
--- a/api/analyze.ts
+++ b/api/analyze.ts
@@ -32,6 +32,29 @@ function getFirestoreDatabaseId(): string {
   );
 }
 
+// Strip path separators and traversal segments from client-supplied filenames
+// before they become part of a Storage object path. Without this, a raw
+// filename such as "team/Q3.pdf" or "report_.._final.pdf" produces an object
+// path that the download guard will refuse to sign, making the file
+// permanently un-downloadable.
+function sanitizeStorageFilename(filename: string): string {
+  let name =
+    String(filename || "document.pdf").replace(/\\/g, "/").split("/").pop() ||
+    "document.pdf";
+  name = name
+    .replace(/\.\./g, "_")
+    .replace(/[\/\\]/g, "_")
+    .replace(/[\x00-\x1f\x7f]/g, "_")
+    .trim();
+  if (!name || name === "." || name === "..") name = "document.pdf";
+  if (name.length > 120) {
+    const extMatch = name.match(/\.[a-zA-Z0-9]{1,10}$/);
+    const ext = extMatch ? extMatch[0] : "";
+    name = name.slice(0, 120 - ext.length) + ext;
+  }
+  return name;
+}
+
 type AnalysisResponse = {
   summary: string;
   key_metrics: Record<string, unknown>;
@@ -534,7 +557,8 @@ full_report MUST be at least 300 words.`;
         : "low";
     const riskLevel = normalizeRiskLevel(rawRisk);
     const now = new Date();
-    const storagePath = `analyses/${ownerId}/${now.getTime()}_${filename}`;
+    const safeFilename = sanitizeStorageFilename(filename);
+    const storagePath = `analyses/${ownerId}/${now.getTime()}_${safeFilename}`;
     const fileUrl = `https://finsight.local/storage/${encodeURIComponent(storagePath)}`;
 
     const docData: any = {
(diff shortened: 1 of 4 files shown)


ionfwsrijan:

PR #615, opened 2026-08-07
fix: route all server Firestore calls through the configured database id

### LINKED ISSUE #605: Mixed Firestore database usage: server uses the default `(default)` DB for role lookups and health checks while documents are written to a configured named DB
## Severity
Medium (config-dependent data consistency: with a named Firestore DB configured, role lookups and health checks silently read the wrong database)

## Affected code
- `server.ts`
  - `enrichUserContext` — `getFirestore()` (line 527)
  - `/api/health` — `getFirestore()` (line 730)
  - Analyze handler — `getFirestore(firestoreDatabaseId)` (line 1195)
  - Signed-URL handler — `getFirestore(firestoreDatabaseId)` (line 1405)
  - Delete handler — `getFirestore(firestoreDatabaseId)` (line 1497)
- `api/process.ts`, `api/analyze.ts`
  - Serverless persistence uses `getFirestore(firestoreDatabaseId)` too

## Description

The server initializes Firestore through a helper (`getFirestore`) that switches on `FIREBASE_FIRESTORE_DATABASE_ID` (`server.ts` import area), and the storage rules documentation already warns that the named database changes which `storage.rules` are enforced. However, three call sites bypass the named DB and use the **default** `getFirestore()`:

```ts
// server.ts:527 (enrichUserContext)
const db = getFirestore();
const userRef = db.collection("users").doc(req.user.uid);
```

```ts
// server.ts:730 (/api/health)
await getFirestore().getAll(...);
```

When `FIREBASE_FIRESTORE_DATABASE_ID` is set (the deployment this repo documents), all user data lives in the named database, but:

1. `enrichUserContext` looks up `users/{uid}` in the **default** database. The lookup always misses, so `req.userRole` stays `undefined`, and the `analyzeRateLimiter` premium/admin skip (`server.ts:531-546`) never activates — every user is rate limited as a free user.
2. `/api/health` checks collection availability in the wrong database and can report "healthy" (or fail) based on collections that don't exist in the actual production database.
3. The data the server sees is fragmented across two databases depending on which handler runs, so any future server-side cross-collection logic reading `users` (e.g. admin gating) will silently operate on empty results.

This is purely configuration-dependent (default-DB-only deployments are unaffected), which is why it has not surfaced in local dev, but it is a latent correctness and access-control bug for every named-DB production deployment.

## Reproduction

1. Set `FIREBASE_FIRESTORE_DATABASE_ID=prod` and populate `users`/`documents` in the `prod` database.
2. Start the server; send an authenticated request to `/api/analyze`.
3. Observe `req.userRole` is always `undefined` (role read from empty default DB) while `documents` reads/writes in `enrichUserContext`'s sibling handlers hit the `prod` DB. `/api/health` reflects the default DB, not `prod`.

## Suggested fix

- Route every Firestore call through the same `getFirestore(firestoreDatabaseId)` helper; remove bare `getFirestore()` call sites at `server.ts:527` and `server.ts:730`.
- Add an integration test that asserts role lookups and document writes land in the same database.

ionfwsrijan:
## Problem

The server mixed Firestore database usage: document writes, signed-URL checks, and deletion all used `getFirestore(firestoreDatabaseId)`, but two call sites used the bare default database:

- `enrichUserContext` (server.ts) looked up `users/{uid}` with `getFirestore()` — with a named database configured, the role lookup always missed, so `req.userRole` stayed `undefined` and the premium/admin rate-limit skip never activated.
- `/api/health` checked collection availability with `getFirestore()` — it could report healthy (or fail) based on the wrong database.
- `api/process.ts` also resolved Firestore through `admin.firestore()` (default DB) while `api/analyze.ts` used the configured database id.

## Fix

- `server.ts` — `enrichUserContext` and `/api/health` now use `getFirestore(firestoreDatabaseId)`, matching the analyze, signed-URL, and delete handlers.
- `api/process.ts` — `getAdminApp()` now resolves Firestore via `admin.firestore(getFirestoreDatabaseId())`.
- Added `tests/firestore-database-consistency.test.mjs` asserting there are no bare `getFirestore()` call sites in `server.ts`, that the named-id calls exist across the handlers, and that both serverless handlers resolve the configured database.

## Files changed

- `server.ts`
- `api/process.ts`
- `tests/firestore-database-consistency.test.mjs`

## Testing

- `node --test tests/firestore-database-consistency.test.mjs` passes (3 tests).
- Full static suite: `tests/storage-rules.test.mjs` and `tests/analyze-quota-guard.test.mjs` still pass. `tests/anomaly-leave-one-out.test.mjs` and `tests/bill-payment-lifecycle.test.mjs` fail only because `node_modules` (e.g. `date-fns`) is not installed in the working copy — pre-existing, unrelated.
- `npm run typecheck` / `npm run lint` could not be executed locally for the same reason.

Closes #605

### DIFF
--- a/api/process.ts
+++ b/api/process.ts
@@ -373,7 +373,10 @@ async function getAdminApp(): Promise<any | null> {
       }
     }
 
-    _adminApp = { admin, getFirestore: () => admin.firestore() };
+    _adminApp = {
+      admin,
+      getFirestore: () => admin.firestore(getFirestoreDatabaseId()),
+    };
     return _adminApp;
   } catch (err: any) {
     console.warn("[process] Firebase Admin init failed:", err?.message);
--- a/server.ts
+++ b/server.ts
@@ -524,7 +524,7 @@ async function enrichUserContext(req: any, res: any, next: any) {
       return next();
     }
 
-    const db = getFirestore();
+    const db = getFirestore(firestoreDatabaseId);
     const userDoc = await db.collection("users").doc(req.ownerId).get();
     req.userRole = userDoc.data()?.role || "free";
     next();
@@ -727,7 +727,7 @@ async function startServer() {
 
     // Check Firestore connectivity
     try {
-      const db = getFirestore();
+      const db = getFirestore(firestoreDatabaseId);
       await db.listCollections();
       checks.firestore = "ok";
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/tests/firestore-database-consistency.test.mjs
+++ b/tests/firestore-database-consistency.test.mjs
@@ -1,0 +1,49 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+
+const repoRoot = path.resolve(path.dirname(fileURLToPath(import.meta.url)), "..");
+const serverTs = readFileSync(path.join(repoRoot, "server.ts"), "utf8").replace(
+  /\r\n/g,
+  "\n",
+);
+const processApi = readFileSync(
+  path.join(repoRoot, "api", "process.ts"),
+  "utf8",
+).replace(/\r\n/g, "\n");
+const analyzeApi = readFileSync(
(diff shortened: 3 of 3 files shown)


ionfwsrijan:

PR #617, opened 2026-08-07
fix: exclude soft-deleted bills from bill surfaces and reminders (Issue #607)

### LINKED ISSUE #607: Soft-deleted bills are still returned by `fetchUserBills` and keep driving bill reminders
## Severity
Medium (functional: "deleted" bills remain visible and keep firing due/overdue reminders)

## Affected code
- `src/lib/billUtils.ts`
  - `softDeleteBill` — only sets `deleted: true`, never removes the doc (lines 129-138; flag set at line 131)
  - `fetchUserBills` — queries `bills` by `userId` with no `deleted` filter (lines 67-99)
- `src/components/bills/BillReminders.tsx`
  - Renders bills from `fetchUserBills` without filtering `deleted` (line 94, and the rendering loop below it)
- `src/components/bills/BillManager.tsx`
  - Lists the same unfiltered result set

## Description

Deleting a bill does not delete it — it flips a flag:

```ts
// src/lib/billUtils.ts:129-138
export async function softDeleteBill(billId: string): Promise<boolean> {
  try {
    await updateDoc(doc(db, 'bills', billId), { deleted: true });
    return true;
  } catch (error) { ... }
}
```

But the query that feeds every bill surface never filters on that flag:

```ts
// src/lib/billUtils.ts:67-99
const q = query(
  billsRef,
  where('userId', '==', userId),
  orderBy('dueDate', 'asc')
);
```

`BillReminders.tsx:94` calls `fetchUserBills(user.uid)` and renders the result set directly (with no client-side `deleted` filter anywhere in the file), and `BillManager.tsx` renders the same list. Consequences:

1. A deleted bill still appears in the bill list and the reminders panel.
2. The reminder engine keeps computing due/overdue notifications for it, because the underlying document remains present and active — the `deleted` field is effectively write-only.
3. Deletion gives users the impression the bill is gone while it continues driving alerts and appears in exports/queries.

## Reproduction

1. Create a recurring bill due tomorrow.
2. Delete it via the UI (`softDeleteBill` sets `deleted: true`).
3. Open the reminders panel — the bill still appears as due/overdue and keeps generating reminders; it also still shows in `BillManager`.

## Suggested fix

- Filter on `deleted` in `fetchUserBills` (e.g. `where('deleted', '==', false)`, or exclude `data.deleted === true` in the mapping loop at `billUtils.ts:77-92`).
- Ensure `markBillAsPaid` and any reminder queries use the same predicate so deleted bills stop driving notifications.

ionfwsrijan:
## Problem

`softDeleteBill` flips `deleted: true` on the bill document instead of removing it, but `fetchUserBills` queries `bills` by `userId` with no `deleted` filter. Every bill surface (`BillReminders`, `BillCard`) renders that unfiltered result set, so a "deleted" bill:

- still appears in the bill list and reminders panel,
- keeps firing due/overdue reminders (`isOverdue`/`isUpcoming` only checked `isPaid`),
- still counts toward monthly obligations.

The `deleted` flag was effectively write-only.

## Fix

- Added `deleted?: boolean` to the `Bill` interface.
- `fetchUserBills` now skips documents with `data.deleted === true` in the mapping loop (safe for legacy bills that predate the flag) and maps `deleted: data.deleted === true` onto the result.
- Defense in depth for any caller that passes deleted bills directly:
  - `isOverdue` and `isUpcoming` return `false` for deleted bills, so `getOverdueBills`/`getUpcomingBills` and reminder checks stop including them.
  - `calculateMonthlyObligations` excludes deleted bills.
  - `markBillAsPaid` refuses to process a deleted bill.
- `BillReminders`' `activeBills` list now filters `!b.deleted` as well.

## Files changed

- `src/lib/billUtils.ts`
- `src/components/bills/BillReminders.tsx`
- `tests/soft-deleted-bills.test.mjs` (new)

## Testing

- Added `tests/soft-deleted-bills.test.mjs` (3 tests) asserting the fetch loop skips/maps `deleted`, reminder/obligation helpers ignore deleted bills, and the active list excludes them.
- `node --test tests/soft-deleted-bills.test.mjs` and the other static suites pass.
- Note: `tests/bill-payment-lifecycle.test.mjs` still cannot run in this checkout because `node_modules` (date-fns) is not installed — a pre-existing limitation, unrelated to this change.

Closes #607

### DIFF
--- a/src/components/bills/BillReminders.tsx
+++ b/src/components/bills/BillReminders.tsx
@@ -118,7 +118,7 @@ export function BillReminders({ user }: BillRemindersProps) {
 
   const overdueBills = useMemo(() => getOverdueBills(bills, today), [bills, today]);
   const upcomingBills = useMemo(() => getUpcomingBills(bills, today), [bills, today]);
-  const activeBills = useMemo(() => bills.filter((b) => !b.isPaid), [bills]);
+  const activeBills = useMemo(() => bills.filter((b) => !b.isPaid && !b.deleted), [bills]);
   const monthlyObligations = useMemo(() => calculateMonthlyObligations(bills), [bills]);
 
   const scheduleData = useMemo(() => {
--- a/src/lib/billUtils.ts
+++ b/src/lib/billUtils.ts
@@ -41,6 +41,7 @@ export interface Bill {
   lastPaidDate?: string | null;
   createdAt: string;
   nextDueDate?: string | null;
+  deleted?: boolean;
 }
 
 export interface BillInput {
@@ -76,6 +77,7 @@ export async function fetchUserBills(userId: string): Promise<Bill[]> {
     const bills: Bill[] = [];
     snapshot.forEach((docSnap) => {
       const data = docSnap.data();
+      if (data.deleted === true) return;
       bills.push({
         id: docSnap.id,
         userId: data.userId || '',
@@ -88,6 +90,7 @@ export async function fetchUserBills(userId: string): Promise<Bill[]> {
         lastPaidDate: data.lastPaidDate || null,
         createdAt: data.createdAt || '',
         nextDueDate: data.nextDueDate || null,
+        deleted: data.deleted === true,
       });
     });
     return bills;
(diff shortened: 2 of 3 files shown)


ionfwsrijan:

PR #618, opened 2026-08-07
fix: load privacy settings from Firestore with localStorage as offline cache (Issue #608)

### LINKED ISSUE #608: Privacy settings are loaded from localStorage but written to Firestore — saved settings are ignored after a cache clear or on another device
## Severity
Medium (data-loss/consistency: user privacy choices silently reset)

## Affected code
- `src/components/privacy/PrivacyDashboard.tsx`
  - Settings load — reads only localStorage (lines 43-64)
  - `handleSettingUpdate` — persists via `updatePrivacySettings` to Firestore (lines 71-74)
  - `getPrivacySettings` imported but never called (line 19)
- `src/lib/privacyUtils.ts`
  - `updatePrivacySettings` — writes to `users`/`privacy_settings/{uid}` in Firestore (lines 73-86)
  - `getPrivacySettings` — reads from Firestore (lines 58-71)

## Description

The privacy panel writes the user's preferences to **Firestore** (`updatePrivacySettings`, `privacyUtils.ts:73-86`) but initializes its state only from **localStorage**:

```ts
// PrivacyDashboard.tsx:43-64
const stored = localStorage.getItem(`privacySettings_${user.uid}`);
setPrivacySettings(stored ? JSON.parse(stored) : { ...defaults });
```

`getPrivacySettings` (which reads the Firestore copy, `privacyUtils.ts:58-71`) is imported at `PrivacyDashboard.tsx:19` but never invoked anywhere in the component. Consequences:

1. On a fresh browser profile, private window, or after `clearSiteData`, the panel re-initializes to `DEFAULT_PRIVACY_SETTINGS` even though the user previously saved preferences to Firestore — the saved choices are silently discarded.
2. Because preferences are also written to localStorage on change, a stale local copy can overwrite newer Firestore values on the next toggle, making the Firestore record drift from what the user actually selected.
3. Multiple devices disagree: each device uses its own localStorage snapshot; there is no single source of truth.

For a privacy feature this is especially harmful: a user who disabled a sharing/export option sees it silently re-enabled after a cache clear.

## Reproduction

1. Open PrivacyDashboard, toggle "Share anonymized data" off, save — both localStorage and Firestore now store `false`.
2. Clear site data for the origin (or open a fresh profile).
3. Reopen PrivacyDashboard — the toggle is back on (default), because load reads only localStorage and never calls `getPrivacySettings`.

## Suggested fix

- Load settings from `getPrivacySettings()` (Firestore) on mount and fall back to localStorage only as an offline cache; write through to Firestore on every change.
- Remove the localStorage-first initialization or reconcile local cache with the Firestore snapshot on load.

ionfwsrijan:
## Problem

The privacy panel writes preferences to Firestore (`privacy_settings/{uid}`) but initializes its state only from localStorage. `getPrivacySettings` (the Firestore reader) was imported but never called. Consequences:

- After a cache clear, private window, or new device, the panel re-initializes to defaults, silently discarding the user's saved privacy choices.
- A stale local copy can overwrite newer Firestore values on the next toggle, so the Firestore record drifts from what the user actually selected.
- No single source of truth across devices — especially harmful for a privacy feature (e.g. a user who disabled data sharing sees it silently re-enabled).

## Fix

- `getPrivacySettings` now merges the stored document over `DEFAULT_PRIVACY_SETTINGS` (so documents written before every field existed still return a complete typed object), and **re-throws** on a read failure instead of silently returning defaults — that lets the dashboard fall back to its cache rather than presenting defaults that would overwrite saved choices.
- `PrivacyDashboard.loadPrivacyData` now loads from `getPrivacySettings(user.uid)` (Firestore) on mount as the single source of truth, and only falls back to the localStorage cache when the Firestore read fails (offline case). The existing effect still writes through to Firestore and syncs the cache on every change, so all three requirements hold:
  - Firestore is authoritative,
  - localStorage is an offline cache only,
  - every change is written through to Firestore.

## Files changed

- `src/lib/privacyUtils.ts`
- `src/components/privacy/PrivacyDashboard.tsx`
- `tests/privacy-settings-source-of-truth.test.mjs` (new)

## Testing

- Added `tests/privacy-settings-source-of-truth.test.mjs` (2 tests) asserting the dashboard loads from Firestore with a cache fallback and that `getPrivacySettings` re-throws on failure instead of returning defaults.
- `node --test tests/privacy-settings-source-of-truth.test.mjs` and the other static suites pass.
- Note: new users now see the Firestore-seeded defaults (`dataRetentionEnabled`/`analyticsEnabled` `false`) instead of the old localStorage-only `true`/`true` — this is the point of the fix (consistent, single source of truth).
- Could not run `npm run typecheck` / `npm run lint`: `node_modules` is not installed in this checkout.

Closes #608

### DIFF
--- a/src/components/privacy/PrivacyDashboard.tsx
+++ b/src/components/privacy/PrivacyDashboard.tsx
@@ -18,10 +18,12 @@ import {
   deleteUserData,
   getPrivacySettings,
   updatePrivacySettings,
+  DEFAULT_PRIVACY_SETTINGS,
   PrivacySettings,
 } from "@/src/lib/privacyUtils";
 import { toast } from "sonner";
 import { auth } from "@/src/lib/firebase";
+import { safeJsonParse } from "@/src/lib/utils";
 import { deleteUser, signOut } from "firebase/auth";
 
 export function PrivacyDashboard({ user }: { user: any }) {
@@ -44,19 +46,19 @@ export function PrivacyDashboard({ user }: { user: any }) {
     if (!user) return;
      
     setIsLoading(true);
+    const cached = localStorage.getItem(`privacySettings_${user.uid}`);
+    const cachedSettings = cached
+      ? (safeJsonParse(cached, null as Partial<PrivacySettings> | null))
+      : null;
     try {
-      const stored = localStorage.getItem(`privacySettings_${user.uid}`);
-      setPrivacySettings(stored ? JSON.parse(stored) : {
-        userId: user.uid,
-        dataRetentionEnabled: true,
-        analyticsEnabled: true,
-        sharingEnabled: false,
-        exportRequestedAt: "",
-        deletionRequestedAt: "",
-        updatedAt: new Date().toISOString(),
-      });
+      // Firestore is the single source of truth. localStorage is only an
+      // offline cache so saved choices survive neither cache clears nor
+      // cross-device switches.
+      const remote = await getPrivacySettings(user.uid);
+      setPrivacySettings({ ...DEFAULT_PRIVACY_SETTINGS, ...remote });
     } catch (error) {
       console.error("Failed to load privacy data:", error);
+      setPrivacySettings({ ...DEFAULT_PRIVACY_SETTINGS, ...cachedSettings });
     } finally {
        
       setIsLoading(false);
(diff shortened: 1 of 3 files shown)


ionfwsrijan:

PR #619, opened 2026-08-07
fix: delete uploaded PDF on Firestore permission-denied fallback (Issue #609)

### LINKED ISSUE #609: PDFs uploaded on the Express path are kept in Storage when Firestore rejects the write — an unreachable, un-deletable orphaned object
## Severity
Medium (resource leak + data hygiene: every permission-denied fallback leaves a permanent orphaned PDF that can never be downloaded or cleaned up)

## Affected code
- `server.ts`
  - PDF uploaded to Storage before the Firestore metadata write (lines 1139-1165)
  - Permission-denied fallback returns a local record and keeps the object (lines 1233-1245)
  - Object deletion happens only on non-permission-denied write failures (lines 1246-1264)
- `server.ts`
  - `POST /api/document-download-url` refuses to sign files without a live owning Firestore record (lines 1401-1422)
- `src/lib/privacyUtils.ts`
  - Account deletion never sweeps orphaned objects (lines 204-267)

## Description

The Express analyze pipeline uploads the PDF to Firebase Storage **before** writing the Firestore document (intentionally, per the comment at `server.ts:1141`). If the subsequent Firestore write fails, the behavior depends on the error type:

```ts
// server.ts:1233-1245
const isPermissionDenied = writeErrorCode === "7" || writeErrorCode === "permission-denied" || ... ;
if (isPermissionDenied) {
  documentId = `local-${ownerId}-${now.getTime()}`;
  // ... return a local-only analysis record
} else {
  // ... delete the just-uploaded PDF so no orphaned object remains (1246-1264)
  await bucket.file(storagePath).delete();
  throw new PipelineError("FIRESTORE_WRITE", ...);
}
```

The non-permission-denied path explicitly deletes the PDF "so a failed pipeline does not leave a permanent orphaned object" — but the **permission-denied** path deliberately skips that cleanup and returns `200` with `persistenceMode: "local"` (`server.ts:1273-1284`). The uploaded object is now unreachable from any supported flow:

1. The download endpoint requires a Firestore document whose `storagePath` matches and whose `ownerId` matches (`server.ts:1401-1422`) — no such record exists for the local-only document, so the file can never be downloaded (403).
2. Account deletion (`privacyUtils.ts:204-267`) only deletes Storage prefixes that contain the user id and are backed by known records; an orphaned object with no Firestore record is never enumerated or removed.
3. The object still carries `uploadedBy` metadata and a full financial PDF, persisting in Storage indefinitely with no owner, no record, and no audit trail.

The same asymmetry exists in the serverless handlers: `api/process.ts:486` and `api/analyze.ts:589` fall back to `local-*` ids on any Firestore write error without ever having uploaded a file (see the separate phantom-URL issue), so on those endpoints the storage side is not the problem — here it specifically is.

## Reproduction

1. Deploy with a service account whose credentials lack Firestore write permission (or revoke them), keeping Storage access enabled.
2. Upload a PDF via `FileUpload.tsx` → `/api/process`.
3. The server logs `FIRESTORE_WRITE_FALLBACK`, returns `200` with `persistenceMode: "local"`, and the PDF remains in Storage under `analyses/<uid>/...`.
4. Confirm the object exists in the bucket, cannot be fetched via `POST /api/document-download-url` (403, no owning record), and is not removed by account deletion.

## Suggested fix

- On the permission-denied fallback, delete the just-uploaded object the same way the non-permission-denied path does (or move the upload to after a successful write).
- Add a background sweep / retention job that lists the `analyses/` prefix and removes objects with no matching Firestore `storagePath` record older than N hours.

ionfwsrijan:
## Problem

On the Express analyze pipeline, the PDF is uploaded to Firebase Storage *before* the Firestore metadata write. When Firestore rejects the write:

- the non-permission-denied path deleted the just-uploaded object,
- but the **permission-denied** fallback returned `200` with `persistenceMode: "local"` and kept the object in Storage.

That object is permanently orphaned: the download endpoint refuses to sign files without a live owning Firestore record (`server.ts:1401-1422`), and account deletion only removes storage backed by known records — so the file (with `uploadedBy` metadata and a full financial PDF) is unreachable and undeletable.

## Fix

- Extracted the storage cleanup into a single `cleanupUploadedPdf` helper and now call it on **both** write-failure branches:
  - permission-denied fallback → deletes the uploaded PDF before returning the local record,
  - other write failures → deletes it as before.
- The 404 case (object not found) is still ignored so the helper is safe when no upload actually happened (e.g. when Admin is uninitialized).

## Files changed

- `server.ts`
- `tests/orphaned-upload-cleanup.test.mjs` (new)

## Testing

- Added `tests/orphaned-upload-cleanup.test.mjs` asserting both branches invoke the shared cleanup that deletes `storagePath`.
- `node --test tests/orphaned-upload-cleanup.test.mjs`, `tests/storage-rules.test.mjs`, and `tests/analyze-quota-guard.test.mjs` all pass.
- Could not run `npm run typecheck` / `npm run lint`: `node_modules` is not installed in this checkout.

Follow-up (not included to keep this change minimal): a background sweep/retention job that lists the `analyses/` prefix and removes objects whose `storagePath` has no matching Firestore record older than N hours.

Closes #609

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -1238,15 +1238,11 @@ CRITICAL RULES:
             writeErrorMessage.includes("permission-denied") ||
             writeErrorMessage.includes("missing or insufficient permissions");
 
-          if (isPermissionDenied) {
-            documentId = `local-${ownerId}-${now.getTime()}`;
-            console.warn(
-              "FIRESTORE_WRITE_FALLBACK: returning local analysis record because Firestore writes are not available",
-            );
-          } else {
-            // The PDF was already uploaded to Storage before these writes began.
-            // Delete it so a failed pipeline does not leave a permanent orphaned
-            // object (with uploadedBy metadata) behind.
+          // The PDF was already uploaded to Storage before these writes began.
+          // Delete it on ANY Firestore write failure so a failed pipeline never
+          // leaves a permanent orphaned object (with uploadedBy metadata) that
+          // cannot be downloaded (no owning Firestore record) or swept later.
+          const cleanupUploadedPdf = async () => {
             try {
               const bucket = getStorage().bucket();
               await bucket.file(storagePath).delete();
@@ -1262,6 +1258,16 @@ CRITICAL RULES:
                 );
               }
             }
+          };
+
+          if (isPermissionDenied) {
+            await cleanupUploadedPdf();
+            documentId = `local-${ownerId}-${now.getTime()}`;
+            console.warn(
+              "FIRESTORE_WRITE_FALLBACK: returning local analysis record because Firestore writes are not available",
+            );
+          } else {
+            await cleanupUploadedPdf();
             throw new PipelineError(
               "FIRESTORE_WRITE",
               `Firestore database write failed: ${writeError?.message || String(writeError)}`,
--- a/tests/orphaned-upload-cleanup.test.mjs
+++ b/tests/orphaned-upload-cleanup.test.mjs
@@ -1,0 +1,35 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+
+const repoRoot = path.resolve(path.dirname(fileURLToPath(import.meta.url)), "..");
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #684, opened 2026-08-08
fix: verify Storage ownership before deleting documents (issue #674)

### LINKED ISSUE #674: [Security]: DELETE /api/documents/delete removes any Storage object without verifying ownership - missing uploadedBy check the download path enforces
## Description

`DELETE /api/documents/delete` (server.ts:1520-1587) verifies that the caller owns the **Firestore record** (`docData?.ownerId !== req.ownerId` → 403), but it never verifies that the **Storage object** being deleted belongs to the caller. It reads `storagePath` straight from the record and executes `getStorage().bucket().file(storagePath).delete()` with the Admin SDK, which bypasses `storage.rules` entirely.

The sibling endpoint `/api/document-download-url` was hardened against exactly this pattern: it fetches the object's metadata and rejects unless `metadata.metadata.uploadedBy === req.ownerId` (server.ts:1413-1422). The delete endpoint has no equivalent check.

Because `isValidDocument` in `firestore.rules` (lines 25-33) requires only `keys().hasAll(['ownerId', 'fileName', 'fileUrl', 'status'])` and does **not** restrict extra fields, any signed-in user can create their own `documents/{x}` record with `ownerId: self` and an arbitrary `storagePath`. The ownership gate then passes, and the server permanently deletes the file at that path.

## Expected Behaviour

Deleting a document must delete only storage objects owned by the requesting user (same `uploadedBy` metadata check enforced on the download path).

## Actual Behaviour

- An attacker creates a `documents` record with their own `ownerId` but `storagePath: "analyses/<victimUid>/<epochMs>_<filename>"` (paths are semi-guessable from upload time + filename).
- `POST /api/documents/delete` passes the Firestore-ownership check and calls `bucket.file(storagePath).delete()` on the victim's object.
- The victim's PDF is permanently deleted with no metadata check and no mismatch log. The Firestore rules cannot defend here because the delete runs server-side with the Admin SDK.

## Proposed Fix

Before deleting, fetch `bucket.file(storagePath).getMetadata()` and require `metadata?.metadata?.uploadedBy === req.ownerId` — mirroring the download endpoint (server.ts:1413-1422). Also validate that the resolved path still lives under the caller's own prefix (e.g. `analyses/${req.ownerId}/`).

## Affected Files

- `server.ts` (`/api/documents/delete` handler, ~lines 1520-1587)
- `firestore.rules` (`isValidDocument`, lines 25-33)

## Impact

**High (security)** — any authenticated user can delete another user's uploaded financial PDFs, causing unrecoverable data loss of financial documents.

ionfwsrijan:
## Problem

`DELETE /api/documents/delete` verified Firestore-record ownership but never verified that the Storage object being deleted belonged to the caller. `isValidDocument` in `firestore.rules` did not restrict extra fields, so any signed-in user could create a `documents` record with their own `ownerId` and an arbitrary `storagePath` pointing at another user's PDF. The ownership gate passed and the server permanently deleted the victim's file (Admin SDK bypasses `storage.rules`).

## Fix

- Mirror the download endpoint's guard in the delete handler: before purging, require the resolved `storagePath` to live under `analyses/<ownerId>/` and fetch the object's metadata, rejecting with 403 unless `metadata.metadata.uploadedBy === req.ownerId`. A missing object (404) is tolerated so legitimate cleanups still work; any other metadata failure returns 500 without deleting the record.
- Harden `isValidDocument` in `firestore.rules` so a `documents` record may only be created with a `storagePath` inside the creator's own namespace (`analyses/<uid>/`).

## Files changed

- `server.ts` — `/api/documents/delete` handler now verifies Storage path prefix and `uploadedBy` metadata before deleting.
- `firestore.rules` — `isValidDocument` now restricts `storagePath` to the caller's own namespace.

## Testing

- Manually traced the handler: a forged record with a victim's `storagePath` now hits the path-prefix or `uploadedBy` mismatch and returns 403 before any delete is issued; legitimate records (own prefix + own `uploadedBy`) delete exactly as before.

Closes #674

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -29,7 +29,14 @@ service cloud.firestore {
              // Only allow https fileUrl so a stored link can never render as
              // a javascript:/data: URL or a link-injection vector.
              data.fileUrl is string && data.fileUrl.size() <= 2048 && data.fileUrl.startsWith('https://') &&
-             data.status in ['pending', 'processing', 'completed', 'failed'];
+             data.status in ['pending', 'processing', 'completed', 'failed'] &&
+             // If a storagePath is stored on the record, it must live inside
+             // the caller's own namespace. Otherwise a client could create a
+             // document whose storagePath points at another user's object and
+             // abuse the server-side purge/delete endpoints that trust it.
+             (!data.keys().has('storagePath') ||
+              (data.storagePath is string &&
+               data.storagePath.startsWith('analyses/' + request.auth.uid + '/')));
     }
 
     function isValidAnalysis(data) {
--- a/server.ts
+++ b/server.ts
@@ -1556,6 +1556,64 @@ CRITICAL RULES:
       const storagePath =
         typeof docData?.storagePath === "string" ? docData.storagePath : "";
 
+      // SECURITY: the ownership check above only proves the caller owns the
+      // Firestore record. The record's storagePath is client-controlled at
+      // create time, so it must never be trusted to point at another user's
+      // Storage object. Mirror the download endpoint (see
+      // POST /api/document-download-url, which verifies namespace + uploadedBy
+      // metadata): refuse to delete unless the object lives under the caller's
+      // own prefix AND was uploaded by the caller.
+      if (storagePath) {
+        const expectedPrefix = `analyses/${req.ownerId}/`;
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #685, opened 2026-08-08
fix: remove duplicate firestore.rules match blocks that nullified userId immutability (issue #675)

### LINKED ISSUE #675: [Security]: Duplicate firestore.rules match blocks nullify the userId-immutability check on transactions/anomalies
## Description

`firestore.rules` defines the `transactions` and `anomalies` collections **twice**, and Firestore security rules OR-combine every matching `match` block — a request succeeds if *any* matching block grants it.

- `transactions`: the first block (`match /transactions/{txnId}`) at lines 133-141 has `allow update: if isOwner(existing().userId) || isAdmin();` — **no check that `incoming().userId` stays equal to `existing().userId`**.
- A second block (`match /transactions/{txId}`) at lines 218-223 is stricter: `allow update: if (isOwner(existing().userId) && incoming().userId == existing().userId) || isAdmin();`.

Because the looser first block already grants `update` to the owner, the stricter second block is vacuous — it can never deny anything the first block allowed. The exact same duplicate pattern exists for `anomalies` (lines 116-124 loose, 268-273 strict).

## Expected Behaviour

The `incoming().userId == existing().userId` immutability check should be enforced so users can never transfer or spoof ownership of a transaction/anomaly. The rule author clearly intended this hardening (see the "Per-user RLS" comment block above the second match); the duplicate declaration silently defeats it.

## Actual Behaviour

- A user updates one of their own `transactions`/`anomalies` docs and sets `userId` to a victim's UID.
- The request passes through the looser first match block.
- The victim's queries (`where("userId", "==", victimUid)`) now return the attacker's forged records — cross-user financial data leakage and namespace poisoning of the victim's dashboards/reports.

## Proposed Fix

Remove the duplicate/legacy match blocks (lines 116-124 and 133-141) and keep a single definition per collection that includes the `incoming().userId == existing().userId` condition on `update`.

## Affected Files

- `firestore.rules` (duplicate `match /transactions` at lines 133-141 and 218-223; duplicate `match /anomalies` at lines 116-124 and 268-273)

## Impact

**High (security)** — the intended per-user data-immutability rule is dead code; users can write forged financial records into another user's namespace, leaking and corrupting data.

ionfwsrijan:
## Problem

`firestore.rules` declared the `transactions` and `anomalies` collections twice. Firestore security rules OR-combine every matching `match` block, so the looser legacy block (`allow update: if isOwner(existing().userId) || isAdmin()` with no `incoming().userId` check) granted the update and made the stricter userId-immutability rule dead code. A user could rename the `userId` on their own records to a victim's UID, poisoning the victim's namespace.

## Fix

Remove the duplicate legacy `match /transactions/{txnId}` and `match /anomalies/{anomalyId}` blocks. Each collection now has a single definition whose `update` rule requires `incoming().userId == existing().userId`, so ownership can never be transferred or spoofed.

## Files changed

- `firestore.rules` — deleted the duplicate loose `transactions` and `anomalies` match blocks (20 lines removed); the stricter per-collection rules with the `userId`-immutability check on `update` are now the only definitions.

## Testing

- Rules were reviewed to confirm each of `transactions` and `anomalies` now has exactly one `match` block containing `incoming().userId == existing().userId` on `update`.
- No application code is affected; this is a rules-only change.

Closes #675

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -113,28 +113,8 @@ service cloud.firestore {
       allow delete: if isOwner(existing().ownerId) || isAdmin();
     }
 
-    match /anomalies/{anomalyId} {
-      // SECURITY: Per-user RLS for the anomalies collection
-      // Prevents cross-user access to anomaly detection data
-      allow get: if isOwner(existing().userId);
-      allow list: if request.auth != null && (request.query.userId == request.auth.uid || isAdmin());
-      allow create: if isSignedIn() && isOwner(incoming().userId);
-      allow update: if isOwner(existing().userId) || isAdmin();
-      allow delete: if isOwner(existing().userId) || isAdmin();
-    }
-
     match /subscriptions/{subId} {
       allow read: if isOwner(existing().userId) || isAdmin();
-      allow create: if isSignedIn() && isOwner(incoming().userId);
-      allow update: if isOwner(existing().userId) || isAdmin();
-      allow delete: if isOwner(existing().userId) || isAdmin();
-    }
-
-    match /transactions/{txnId} {
-      // SECURITY: Per-user RLS for the transactions collection
-      // Prevents cross-user access to financial transaction data
-      allow get: if isOwner(existing().userId);
-      allow list: if request.auth != null && (request.query.userId == request.auth.uid || isAdmin());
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #689, opened 2026-08-08
fix: report short flat series as stable in calculateTrend (issue 679)

### LINKED ISSUE #679: [Bug]: calculateTrend can never return 'stable' for 2-3 month series - empty older window yields NaN comparison
## Description

`calculateTrend` in `src/lib/forecastUtils.ts:214-221`:

```ts
export function calculateTrend(data: { month: string; value: number }[]): 'up' | 'down' | 'stable' {
  if (data.length < 2) return 'stable';
  const recent = data.slice(-3).reduce((s, d) => s + d.value, 0) / Math.min(3, data.length);
  const older = data.slice(0, -3).reduce((s, d) => s + d.value, 0) / Math.max(1, data.length - 3);
  const diff = recent - older;
  if (Math.abs(diff) < older * 0.05) return 'stable';
  return diff > 0 ? 'up' : 'down';
}
```

For `data.length === 2` or `3`, `data.slice(0, -3)` is empty, so `older` becomes `0 / 1 = 0`. Then the stability check is `Math.abs(diff) < 0`, which is never true for a nonzero average. The function therefore always returns `'up'` (positive values) or `'down'` (negative values) — a perfectly flat, positive two-month series is labeled "Trending up".

## Expected Behaviour

A flat series should be reported as `'stable'` regardless of how many data points it has.

## Actual Behaviour

`ForecastComparison` (lines 158-161) uses `calculateTrend` to badge the Net Balance card. Any user whose filtered range contains only 2-3 months gets a guaranteed `up`/`down` label with no possible `stable` outcome, even when the values are identical month over month.

## Proposed Fix

When `older` is zero (no older window exists), fall back to comparing against the recent average itself, e.g. guard with `if (older === 0) return Math.abs(diff) < recent * 0.05 ? 'stable' : (diff > 0 ? 'up' : 'down');` or treat 2-3 point series as `'stable'` when values are equal.

## Affected Files

- `src/lib/forecastUtils.ts` (`calculateTrend`, lines 214-221)
- `src/components/forecast/ForecastComparison.tsx` (lines 158-161)

## Impact

**Medium** — the trend indicator is unreliable for short ranges and can never report "stable", misleading users who filter to 2-3 months.

ionfwsrijan:
## Problem

For 2-3 month series `data.slice(0, -3)` is empty, so `older = 0 / 1 = 0`. The stability check `Math.abs(diff) < older * 0.05` then compares against `0` and can never be true for a nonzero average. `ForecastComparison` (lines 158-161) therefore always badges a 2-3 month Net Balance card as `up`/`down` — a perfectly flat series is labeled "Trending up".

## Fix

When `older === 0` (no older window to compare against), classify the available points directly from their endpoints: if `|last - first|` is within 5% of the recent average the series is `'stable'`; otherwise the direction comes from the endpoint change. This matches the existing 5% threshold used for longer series. (The issue's literal proposed one-liner — comparing `diff = recent - 0` against `recent * 0.05` — still returns `up` for a flat positive series, so endpoints are used instead to satisfy the stated expected behaviour.)

Verified cases:

| Series | Result |
| ------ | ------ |
| `[100, 100]`, `[100, 100, 100]`, `[0, 0, 0]`, `[100, 104]` | `stable` |
| `[100, 120]` | `up` |
| `[120, 100]`, `[100, 90, 80]`, `[100, -100]` | `down` |

## Files changed

- `src/lib/forecastUtils.ts` (`calculateTrend`, lines 214-224)

## Testing

- Node-based unit checks of all the edge cases above pass.

Closes #679


ionfwsrijan:

PR #690, opened 2026-08-08
fix: stop overwriting per-challenge difficulty on generation (issue 680)

### LINKED ISSUE #680: [Bug]: Generated challenges have per-challenge difficulty silently overwritten by one global difficulty
## Description

`generateChallenges` in `src/components/challenges/ChallengesDashboard.tsx` (lines 229-245) overwrites every per-challenge difficulty with a single global difficulty:

```ts
const difficulty: Difficulty = calculateDifficulty(spending);   // one global value
const weekly = generateWeeklyChallenges(spending);
const monthly = generateMonthlyChallenges(spending);
const all = [...weekly, ...monthly];
...
for (const data of newChallenges) {
  await createChallenge(user.uid, { ...data, difficulty });   // overrides data.difficulty
  created++;
}
```

But the generators deliberately assign **per-challenge** difficulties: `challengeUtils.ts:186` sets `'Pause One Subscription This Week'` to `difficulty === 'hard' ? 'hard' : 'easy'`, and `challengeUtils.ts:240` sets `'Reach A 20% Savings Rate'` to always `'hard'`. The `{ ...data, difficulty }` spread then unconditionally clobbers those values.

## Expected Behaviour

Each generated challenge should keep the difficulty its generator assigned (the whole point of the per-challenge logic).

## Actual Behaviour

- A user whose global difficulty is `easy` gets "Reach A 20% Savings Rate" stored as `easy` (50 pts) instead of `hard` (400 pts).
- "Pause One Subscription This Week" becomes `medium`/`hard` even for easy spenders.
- The completion toast (`+${DIFFICULTY_REWARDS[challenge.difficulty].points}`, line 201), the difficulty badge, and leaderboard points all reflect the wrong tier.

## Proposed Fix

Drop the override: `await createChallenge(user.uid, data)` so each challenge keeps `data.difficulty`. Use the global `difficulty` only as a fallback for generators that don't set one.

## Affected Files

- `src/components/challenges/ChallengesDashboard.tsx` (line 243, plus toast at 201)
- `src/lib/challengeUtils.ts` (per-challenge difficulty assignment at lines 186, 240)

## Impact

**Medium** — the per-challenge difficulty design is dead code; reward points and difficulty display are wrong for most generated challenges.

ionfwsrijan:
## Problem

`generateChallenges` overwrote every per-challenge difficulty with a single global value (`{ ...data, difficulty }`), clobbering the deliberate per-challenge assignments in `challengeUtils.ts` — e.g. "Reach A 20% Savings Rate" is meant to always be `hard` (400 pts), and "Pause One Subscription This Week" should be `easy`/`hard` based on spend. The completion toast, difficulty badge, and leaderboard points all reflected the wrong tier.

## Fix

Dropped the override: `await createChallenge(user.uid, data)` keeps each challenge's generator-assigned difficulty. `createChallenge` already falls back to `data.difficulty || 'easy'`, so no fallback logic is lost. The now-unused global `difficulty` computation and its `Difficulty`/`calculateDifficulty` imports were removed, and the success toast no longer claims a single global difficulty.

## Files changed

- `src/components/challenges/ChallengesDashboard.tsx` (`generateChallenges`)

## Testing

- All weekly/monthly generators (challengeUtils.ts:151-245) explicitly set `difficulty`, so no challenge loses its intended tier.

Closes #680

### DIFF
--- a/src/components/challenges/ChallengesDashboard.tsx
+++ b/src/components/challenges/ChallengesDashboard.tsx
@@ -27,7 +27,6 @@ import { db, handleFirestoreError, OperationType } from '@/src/lib/firebase';
 import {
   type Challenge,
   type SpendingPattern,
-  type Difficulty,
   type BadgeTier,
   BADGE_META,
   DIFFICULTY_REWARDS,
@@ -35,7 +34,6 @@ import {
   generateWeeklyChallenges,
   generateMonthlyChallenges,
   generateRecommendations,
-  calculateDifficulty,
   awardBadge,
   getProgressPercentage,
   createChallenge,
@@ -226,7 +224,6 @@ export function ChallengesDashboard({ user }: ChallengesDashboardProps) {
     }
     setGenerating(true);
     try {
-      const difficulty: Difficulty = calculateDifficulty(spending);
       const weekly = generateWeeklyChallenges(spending);
       const monthly = generateMonthlyChallenges(spending);
       const all = [...weekly, ...monthly];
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #761, opened 2026-08-10
fix: correct Conservative/Aggressive goal contribution suggestions (issue #745)

### LINKED ISSUE #745: Conservative/Aggressive goal contribution suggestions are swapped
## Summary

The "Conservative" and "Aggressive" contribution suggestions for goals are swapped: by default Conservative proposes a **higher** monthly amount than Aggressive.

## Root cause

`src/lib/goalUtils.ts:63-76`:

```ts
{
  label: 'Conservative',
  amount: Math.ceil(baseMonthly * (options.conservative ? 0.85 : 1.2)),
},
{
  label: 'Aggressive',
  amount: Math.max(1, Math.floor(baseMonthly * (options.aggressive ? 1.2 : 0.85))),
},
```

With no options set (the normal call path): Conservative = `baseMonthly * 1.2` (the larger number) and Aggressive = `baseMonthly * 0.85` (the smaller number). The labels are inverted, and setting the option flags flips them again — contradictory behavior depending on caller.

## Proposed fix

Drop the option-flag branching and make the labels always correct:

```ts
{
  label: 'Recommended',
  amount: baseMonthly,
},
{
  label: 'Conservative',
  amount: Math.ceil(baseMonthly * 0.85),
},
{
  label: 'Aggressive',
  amount: Math.max(1, Math.floor(baseMonthly * 1.2)),
},
```

ionfwsrijan:
## Problem
The Conservative and Aggressive goal contribution suggestions are swapped. The "Aggressive" plan shows the base monthly contribution while the Recommended label shows a higher amount, so the labels and values don't match.

## Fix
- Removed the `options` parameter from `generateContributionSuggestions`.
- Each label is now derived directly from the base amount: Recommended = base, Conservative = `ceil(base * 0.85)`, Aggressive = `max(1, floor(base * 1.2))`.

## Files changed
- `src/lib/goalUtils.ts`

## Testing
- Verified the label/value mapping with self-contained test cases covering all three plans.

Closes #745

### DIFF
--- a/src/lib/goalUtils.ts
+++ b/src/lib/goalUtils.ts
@@ -46,11 +46,7 @@ export function calculateDaysRemaining(deadline: string): number {
 export function generateContributionSuggestions(
   targetAmount: number,
   currentAmount: number,
-  deadline: string,
-  options: {
-    conservative?: boolean;
-    aggressive?: boolean;
-  } = {}
+  deadline: string
 ): { label: string; amount: number }[] {
   const remaining = targetAmount - currentAmount;
   if (remaining <= 0) return [{ label: 'Goal reached', amount: 0 }];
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

ionfwsrijan:

PR #762, opened 2026-08-10
fix: project goal timeline until target instead of capping at 24 months (issue #746)

### LINKED ISSUE #746: Goal timeline projection hard-capped at 24 months never shows completion for longer goals
## Summary

The goal timeline projection stops at 24 months and never shows the goal being reached for goals that take longer than two years.

## Root cause

`src/lib/goalUtils.ts:92-99`:

```ts
for (let i = 1; i <= 24 && running < targetAmount; i++) {
  running += monthlyContribution;
  if (running > targetAmount) running = targetAmount;
  projection.push({
    month: format(addMonths(now, i), 'MMM yyyy'),
    projected: Math.round(running),
  });
}
```

The loop is hard-capped at `i <= 24`. A goal that needs 30 months returns only 24 points with `running < targetAmount`, so the chart/table silently ends mid-goal with no completion month. (Note `running` is also never rounded at the end and stops at `targetAmount`, so the terminal point is also unrounded.)

## Proposed fix

Loop until the goal is reached, with a generous safety bound:

```ts
for (let i = 1; running < targetAmount && i <= 600; i++) {
  running = Math.min(targetAmount, running + monthlyContribution);
  projection.push({
    month: format(addMonths(now, i), 'MMM yyyy'),
    projected: Math.round(running),
  });
}
```

ionfwsrijan:
## Problem
The goal timeline projection stops after 24 months even when the target amount has not been reached, so long-term goals are shown as never reaching their target.

## Fix
- The projection loop now runs until `running >= targetAmount` with a safety cap of 600 iterations to prevent runaway loops.
- Each running total is clamped with `Math.min(targetAmount, ...)` so the final entry lands exactly on the target instead of overshooting.

## Files changed
- `src/lib/goalUtils.ts`

## Testing
- Verified the loop exits at the target and the cap prevents infinite iteration for very large goals.

Closes #746

### DIFF
--- a/src/lib/goalUtils.ts
+++ b/src/lib/goalUtils.ts
@@ -89,9 +89,8 @@ export function generateTimelineProjection(
   const projection: { month: string; projected: number }[] = [];
   let running = currentAmount;
 
-  for (let i = 1; i <= 24 && running < targetAmount; i++) {
-    running += monthlyContribution;
-    if (running > targetAmount) running = targetAmount;
+  for (let i = 1; running < targetAmount && i <= 600; i++) {
+    running = Math.min(targetAmount, running + monthlyContribution);
     projection.push({
       month: format(addMonths(now, i), 'MMM yyyy'),
       projected: Math.round(running),


ionfwsrijan:

ionfwsrijan:

PR #763, opened 2026-08-10
fix: detect new-category spikes and scale spike delta to category history (issue #747)

### LINKED ISSUE #747: detectCategorySpikes: flat $500 gate plus new categories are never flagged
## Summary

`detectCategorySpikes` requires a flat **$500** month-over-month delta and ignores categories that have no prior baseline, so small spenders and brand-new spending categories are never flagged.

## Root cause

`src/lib/anomalyUtils.ts:166-181`:

```ts
.map(([category, items]) => {
  const categoryBaseline = baseline.get(category);
  ...
  return categoryBaseline
    ? { category, amount, baseline: categoryBaseline, transactions: items }
    : null; // categories with no baseline are dropped
})
.filter((item) => {
  if (!item || item.baseline.monthlyTotals.length < 2) return false;
  ...
  return average > 0 && item.amount > average * 1.5 && item.amount - average > 500;
});
```

Two problems:
1. Line 180: `item.amount - average > 500` — a user whose average category spend is $100 and spikes 2.6× to $260 is below the flat $500 delta and never flagged.
2. Lines 170-172: categories absent from the baseline (i.e. a **new** category with spending this month) are mapped to `null` and filtered out entirely.

## Proposed fix

- Replace the flat `$500` delta with a proportional check (e.g. `> $20` minimum + the existing `1.5×` ratio), so small spenders are caught:
  ```ts
  return average > 0
    && item.amount > average * 1.5
    && item.amount - average > Math.max(20, average * 0.5);
  ```
- For categories with no baseline, flag them when their current-month spend exceeds a sanity floor, e.g. a multiple of the overall per-category average across all categories (`averageAllCategories * 2`), so genuinely new spending is surfaced.

ionfwsrijan:
## Problem
`detectCategorySpikes` silently drops brand-new categories that have no prior-period baseline, and its flat `$500` delta gate misses meaningful smaller spikes while flagging noise.

## Fix
- `detectCategorySpikes` now returns `baseline: CategoryBaseline | null`.
- Brand-new categories (no prior-period spending) are flagged when their amount exceeds 2x the average across all categories.
- Existing categories are flagged when `amount - average > Math.max(20, average * 0.5)`.
- `AnomalyDashboard.tsx` guards for a null baseline when computing confidence, percent over, and the description.

## Files changed
- `src/lib/anomalyUtils.ts`
- `src/components/anomaly/AnomalyDashboard.tsx`

## Testing
- Verified new-category and existing-category spike detection with self-contained test cases.

Closes #747

### DIFF
--- a/src/components/anomaly/AnomalyDashboard.tsx
+++ b/src/components/anomaly/AnomalyDashboard.tsx
@@ -383,17 +383,19 @@ async function runDetection(userId: string) {
   });
 
   categorySpikeAnomalies.forEach((spike) => {
+    const monthlyTotals = spike.baseline?.monthlyTotals ?? [];
     const confidence = calculateConfidenceScore(
       "category_spike",
       spike.amount,
-      spike.baseline.mean,
-      spike.baseline.stdDev,
+      spike.baseline?.mean ?? spike.amount,
+      spike.baseline?.stdDev ?? 0,
     );
     const lastMonthTotal =
-      spike.baseline.monthlyTotals[spike.baseline.monthlyTotals.length - 1];
+      monthlyTotals[monthlyTotals.length - 1] ?? spike.amount;
     const avgMonthly =
-      spike.baseline.monthlyTotals.reduce((a, b) => a + b, 0) /
-      spike.baseline.monthlyTotals.length;
+      monthlyTotals.length > 0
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

ionfwsrijan:

PR #764, opened 2026-08-10
fix: scale category_spike delta threshold to category history (issue #748)

### LINKED ISSUE #748: Month-over-month category_spike anomalies require a >$5000 delta
## Summary

The month-over-month `category_spike` anomaly requires a **> $5,000** absolute delta, so realistic category spikes (e.g. Utilities $600 → $1,500, a 2.5× jump) are never reported.

## Root cause

`src/lib/anomalyUtils.ts:286-319` (inside the detection run):

```ts
thisMonthData.forEach((amount, cat) => {
  const lastAmount = lastMonthData.get(cat) || 0;
  if (
    lastAmount > 0 &&
    amount > lastAmount * 1.5 &&
    amount - lastAmount > 5000     // <-- flat $5,000 gate
  ) { ... }
});
```

The 1.5× ratio check is reasonable, but the `amount - lastAmount > 5000` condition effectively disables the detector for anyone whose category spend is below five figures.

## Proposed fix

Make the delta proportional to the category's own scale, e.g. flag when the increase exceeds 50% of the previous month **and** a small absolute floor:

```ts
const minDelta = Math.max(50, lastAmount * 0.5);
if (
  lastAmount > 0 &&
  amount > lastAmount * 1.5 &&
  amount - lastAmount > minDelta
) { ... }
```

ionfwsrijan:
## Problem
A `category_spike` anomaly requires a delta of more than $5,000 (`lastAmount - average > 5000`), so most legitimate spikes are never reported regardless of the category's normal spend level.

## Fix
- Replaced the flat `> 5000` gate with a threshold derived from the category's own history: `minDelta = Math.max(50, lastAmount * 0.5)`.

## Files changed
- `src/lib/anomalyUtils.ts`

## Testing
- Verified spikes are now flagged proportionally (e.g., a small category with a >50% jump, and a large category with a >50% jump) and flat thresholds no longer suppress valid spikes.

Closes #748

### DIFF
--- a/src/lib/anomalyUtils.ts
+++ b/src/lib/anomalyUtils.ts
@@ -286,10 +286,11 @@ export function detectAnomalies(
   if (thisMonthData && lastMonthData) {
     thisMonthData.forEach((amount, cat) => {
       const lastAmount = lastMonthData.get(cat) || 0;
+      const minDelta = Math.max(50, lastAmount * 0.5);
       if (
         lastAmount > 0 &&
         amount > lastAmount * 1.5 &&
-        amount - lastAmount > 5000
+        amount - lastAmount > minDelta
       ) {
         anomalies.push({
           userId: transactions[0]?.userId || "",


ionfwsrijan:

ionfwsrijan:

PR #765, opened 2026-08-10
fix: return real transaction id for portfolio buy operations (issue #749)

### LINKED ISSUE #749: addTransaction returns a phantom id for buy transactions
## Summary

`addTransaction` returns a **phantom id** for `buy` transactions — the id belongs to a document that was never written — so any caller that updates/deletes/refetches by the returned id targets a non-existent doc.

## Root cause

`src/lib/portfolioUtils.ts:215-318`:

```ts
const id = doc(collection(db, 'portfolios')).id;   // <-- wrong collection

if (input.type === 'buy') {
  await addDoc(collection(db, 'portfolioTransactions'), { ... });  // Firestore assigns a REAL id
} else {
  await setDoc(doc(db, 'portfolioTransactions', id), { ... });     // uses the generated id
}

return { ...transaction, id: id || '' };   // buys return the wrong id
```

For buys, `addDoc` generates its own id that is never captured; the function returns `id` generated from the `portfolios` collection (unrelated). The returned `Transaction.id` therefore does not match the persisted document.

## Proposed fix

Use the id of the actual write:

```ts
if (input.type === 'buy') {
  const ref = await addDoc(collection(db, 'portfolioTransactions'), {
    ...transaction,
    createdAt: serverTimestamp(),
  });
  return { ...transaction, id: ref.id };
}

await setDoc(doc(db, 'portfolioTransactions', id), {
  ...transaction,
  createdAt: serverTimestamp(),
});
return { ...transaction, id };
```

Also generate `id` from `portfolioTransactions` (not `portfolios`) in the non-buy path.

ionfwsrijan:
## Problem
For buy transactions, `addTransaction` returns a phantom id that was generated from the wrong collection, so callers can't reference the transaction that was actually saved.

## Fix
- The transaction id is now generated from the `portfolioTransactions` collection so it matches the collection the document is written to.
- The buy path captures the real `ref.id` returned by `addDoc` and returns it.

## Files changed
- `src/lib/portfolioUtils.ts`

## Testing
- Verified the returned id for buy transactions matches the document created in the `portfolioTransactions` collection.

Closes #749


ionfwsrijan:

PR #766, opened 2026-08-10
fix: exclude outlier from its own anomaly baseline (issue #750)

### LINKED ISSUE #750: detectAnomalies includes the outlier in its own baseline (under-detects)
## Summary

`detectAnomalies` includes each candidate transaction in the baseline used to judge it, so genuine anomalies are frequently under-detected — and the same transaction is re-flagged on every recompute.

## Root cause

`src/lib/insightsUtils.ts:230-238`:

```ts
for (const [category, list] of byCategory.entries()) {
  if (list.length < 3) continue;
  const avg = total(list) / list.length;   // includes the outlier itself
  ...
  for (const tx of list) {
    const ratio = amount / avg;
    if (ratio >= threshold) { ... }
  }
}
```

Example: category list `[100, 100, 250]`. `avg = 150`, so the $250 charge scores `250/150 = 1.67 < 2` and is **not** flagged — even though it is 2.5× the normal $100. The doc comment (lines 211-215) explicitly promises "> threshold x the average spend for their category", which implies the baseline should exclude the transaction.

Additionally, since every call recomputes from raw transactions, already-acknowledged anomalies get flagged again each time (no dismissal state is consulted).

## Proposed fix

Use a leave-one-out average for each transaction:

```ts
for (const tx of list) {
  const othersTotal = total(list) - normalizeAmount(tx.amount);
  const avg = othersTotal / (list.length - 1);
  const ratio = normalizeAmount(tx.amount) / avg;
  if (avg > 0 && ratio >= threshold) { ... }
}
```

And pass previously-dismissed anomaly ids (or a date watermark) so repeat flags are suppressed.

ionfwsrijan:
## Problem
`detectAnomalies` includes the anomalous amount in its own baseline average, which dilutes the outlier so the reported delta and percentage are understated.

## Fix
- Each transaction's baseline now uses a leave-one-out average: `(total - amount) / (count - 1)`, excluding the transaction itself.
- `buildInsights` accepts an optional `ignoredTransactionIds` set so dismissed anomalies aren't fed back into detection on subsequent runs.
- `InsightsDashboard` fetches the user's dismissed anomalies and passes their transaction ids to `buildInsights`.

## Files changed
- `src/lib/insightsUtils.ts`
- `src/components/insights/InsightsDashboard.tsx`

## Testing
- Verified the leave-one-out average isolates the outlier with self-contained test cases (existing `tests/anomaly-leave-one-out.test.mjs` still passes).

Closes #750

### DIFF
--- a/src/components/insights/InsightsDashboard.tsx
+++ b/src/components/insights/InsightsDashboard.tsx
@@ -51,6 +51,7 @@ import {
 } from "@/src/lib/insightsUtils";
 import { handleFirestoreError, OperationType } from "@/src/lib/firebase";
 import { useBaseCurrency } from "@/src/hooks/useBaseCurrency";
+import { fetchAnomalies } from "@/src/lib/anomalyUtils";
 import { InsightCard, CategoryDeltaRow } from "./InsightCard";
 
 const LINE_COLORS = ["#818cf8", "#34d399", "#fbbf24", "#f472b6", "#22d3ee"];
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

ionfwsrijan:

PR #767, opened 2026-08-10
fix: do not schedule subscription reminder timers that fire immediately (issue #752)

### LINKED ISSUE #752: Subscription Remind button schedules a setTimeout that fires immediately for distant/overdue renewals
## Summary

The "Remind" button on a subscription schedules a `setTimeout` that fires **immediately** for renewals more than ~24.8 days away (browser timer cap), fires immediately for overdue renewals (negative delay), and is lost entirely when the tab closes.

## Root cause

`src/components/subscriptions/SubscriptionCard.tsx:71-95`:

```ts
setTimeout(
  () => {
    new Notification("Subscription Renewal Reminder", { body: `...` });
  },
  differenceInDays(reminderDate, new Date()) * 24 * 60 * 60 * 1000,
);
toast.success("Reminder set for renewal");
```

- Browsers clamp `setTimeout` delays to ~2^31 ms (~24.8 days); longer delays **overflow and fire immediately** — so a renewal 30 days out reminds the user right now.
- An overdue renewal has a negative `differenceInDays`, so the delay is ≤ 0 and fires instantly too (and "Reminder set" is toasted regardless).
- A page-`setTimeout` reminder is non-persistent: closing the tab cancels it.

## Proposed fix

Use the service worker's `showTrigger`/`schedule` support (or a daily re-schedule that checks `now >= renewal`) instead of a single long timer:

```ts
// PWA-style approach (SW with showTrigger on Chromium):
const reg = await navigator.serviceWorker.ready;
await reg.showNotification("Subscription Renewal Reminder", {
  body: `${subscription.name} renews soon.`,
  tag: `renewal-${subscription.id}`,
  showTrigger: new TimestampTrigger(subscription.nextRenewalDate),
});
```

Fallback for non-supporting browsers: early-return when `differenceInDays <= 0` or `> 24` so you never schedule an instant/impossible delay, and only toast success when a reminder was actually scheduled.

ionfwsrijan:
## Problem
The subscription "Remind me" timer fires immediately: the `setTimeout` delay is computed from the current date even when the renewal is already due or only days away, and very large delays can overflow the 32-bit timeout limit.

## Fix
- `handleSetReminder` now checks notification permission first and returns early with an informative toast when the renewal is today/already due (`daysUntil <= 0`) or more than 24 days out.
- Prefers the service worker's `showNotification` when available (tagged `renewal-${id}`).
- Falls back to a `setTimeout` whose delay is computed with `differenceInMilliseconds`.

## Files changed
- `src/components/subscriptions/SubscriptionCard.tsx`

## Testing
- Verified the guard conditions (due today, >24 days out) return early with a toast instead of scheduling a timer.

Closes #752


ionfwsrijan:

PR #769, opened 2026-08-10
fix: make markBillAsPaid idempotent and disable pay button while processing (issue #754)

### LINKED ISSUE #754: markBillAsPaid is not idempotent - double-click creates duplicate expenses
## Summary

`markBillAsPaid` is not idempotent: clicking "Mark as Paid" twice (double-click, or a slow write) records **two** expense transactions and advances the due date twice.

## Root cause

`src/lib/billUtils.ts:278-312`:

```ts
export async function markBillAsPaid(bill: Bill, userId: string): Promise<Bill | null> {
  if (bill.deleted) return null;   // guards delete, but NOT bill.isPaid
  try {
    ...
    const batch = writeBatch(db);
    batch.update(doc(db, 'bills', bill.id), payment);
    batch.set(doc(collection(db, 'transactions')), transactionData); // new doc each call
    await batch.commit();
    ...
```

The comment at lines 298-300 claims atomicity "prevents … double-charge on retry", but each invocation writes a **fresh** transaction document (auto-id) and applies `applyBillPayment` again, so two sequential clicks = duplicate expense + double advance.

The Pay button (`src/components/bills/BillCard.tsx:154-165`) has no pending/disabled state between click and state refresh, so a double-click is easy.

## Proposed fix

Guard inside the function and disable the button while processing:

```ts
export async function markBillAsPaid(bill: Bill, userId: string): Promise<Bill | null> {
  if (bill.deleted || bill.isPaid) return { ...bill };  // idempotent
  ...
}
```

And in `src/components/bills/BillReminders.tsx` track a `payingId` state, passing `disabled={payingId === bill.id}` down to `BillCard`, clearing it in a `finally` after the awaited call.

ionfwsrijan:
## Problem
`markBillAsPaid` is not idempotent: clicking "Mark as paid" twice (or after a refresh) can create a duplicate expense transaction. The pay button also stays enabled while a payment is processing.

## Fix
- `markBillAsPaid` now returns the bill unchanged when `bill.deleted || bill.isPaid`, preventing duplicate expense writes.
- `BillReminders` tracks a `payingId` state and passes `disabled={payingId === bill.id}` to each `BillCard`, clearing it in a `finally` block.
- `BillCard` accepts an optional `disabled` prop wired to its pay button.

## Files changed
- `src/lib/billUtils.ts`
- `src/components/bills/BillReminders.tsx`
- `src/components/bills/BillCard.tsx`

## Testing
- Verified double-invocation returns the original bill without a second write.

Closes #754


ionfwsrijan:

PR #771, opened 2026-08-10
fix: exclude partial current month from budget suggestion average (issue #756)

### LINKED ISSUE #756: Budget suggestions average in the partial current month
## Summary

`generateBudgetSuggestions` averages the **partial current month** in with the two full months of history, producing unrealistically low suggested budget limits whenever the app is used mid-month.

## Root cause

`src/lib/budgetUtils.ts:360-364`:

```ts
export async function fetchLast3MonthsTransactions(userId: string): Promise<Transaction[]> {
  const now = new Date();
  const startDate = new Date(now.getFullYear(), now.getMonth() - 2, 1);
  return fetchUserTransactionsInRange(userId, startDate, null); // null end => NOW
}
```

The window is `currentMonth-2 … now`, i.e. it includes the current, incomplete month. `generateBudgetSuggestions` (lines 373-421) then treats every month in the window equally:

```ts
const monthlyTotals = Array.from(monthMap.values());
const averageSpending = monthlyTotals.reduce((sum, v) => sum + v, 0) / monthlyTotals.length;
```

If today is the 10th, a category that normally spends $300/month shows ~$100 for the current month, pulling the 3-month average down to ~$233 — and the suggested budget is set accordingly. Confidence scores (lines 405-408) are also inflated because the partial month counts as a "month".

## Proposed fix

Only average complete months. Exclude the current month from the suggestion window:

```ts
export async function fetchLast3MonthsTransactions(userId: string): Promise<Transaction[]> {
  const now = new Date();
  const endDate = new Date(now.getFullYear(), now.getMonth(), 1);       // first of current month
  const startDate = new Date(now.getFullYear(), now.getMonth() - 3, 1); // 3 full months
  return fetchUserTransactionsInRange(userId, startDate, endDate);
}
```

(Or, if current-month data should be included, scale it by `daysInMonth / daysElapsed` before averaging — but excluding it is simpler and avoids partial-month bias in the confidence scores too.)

ionfwsrijan:
## Problem
Budget suggestions average the partial current month as if it were a full month, dragging category averages (and therefore suggested budgets) downward.

## Fix
- `fetchLast3MonthsTransactions` now caps the fetch window at the first of the current month (`endDate`), matching the existing `fetchPreviousMonthTransactions` pattern, so only complete months are included in the average.

## Files changed
- `src/lib/budgetUtils.ts`

## Testing
- Verified the window is now `[first of month-2, first of current month)`; `fetchUserTransactionsInRange` already excludes any transaction on or after the end date.

Closes #756

### DIFF
--- a/src/lib/budgetUtils.ts
+++ b/src/lib/budgetUtils.ts
@@ -360,7 +360,8 @@ async function fetchUserTransactionsInRange(
 export async function fetchLast3MonthsTransactions(userId: string): Promise<Transaction[]> {
   const now = new Date();
   const startDate = new Date(now.getFullYear(), now.getMonth() - 2, 1);
-  return fetchUserTransactionsInRange(userId, startDate, null);
+  const endDate = new Date(now.getFullYear(), now.getMonth(), 1);
+  return fetchUserTransactionsInRange(userId, startDate, endDate);
 }
 
 export async function fetchPreviousMonthTransactions(userId: string): Promise<Transaction[]> {


ionfwsrijan:

PR #928, opened 2026-08-11
fix(Dashboard): guard normalizeConfidence against missing sentiment data (#916)

### LINKED ISSUE #916: fix : guard normalizeConfidence against fabricated 92 fallback for documents without sentiment

`normalizeConfidence` in `src/components/Dashboard.tsx:31-36` currently returns a hardcoded `92` when a document has no usable sentiment score. This is misleading — documents that have no confidence data at all should not inflate the dashboard's average.

**Expected behaviour:** Return `null` or `undefined` when the value is missing, so the average calculation can exclude it rather than count it as a high-confidence document.

1. In `src/components/Dashboard.tsx`, modify `normalizeConfidence` to return `null` (instead of `92`) when `isNaN(n) || n === 0`.
2. In the same file, update the averaging logic that consumes `normalizeConfidence` to filter out `null` values before including them in `totalConfidenceValues`.

- The Dashboard confidence average will no longer be artificially inflated by documents without any sentiment score.
- Users get an accurate view of how much AI-assisted analysis confidence they actually have.
- Downstream confidence-trend charts will not be skewed by fabricated data points.

saidai-bhuvanesh:
## What

Fixes #916

`normalizeConfidence` in `src/components/Dashboard.tsx` returned a numeric fallback when a document had no usable sentiment score. The original implementation coerced missing values to `0`:

```js
function normalizeConfidence(value: any) {
  const n = Number(value || 0);
  if (n <= 1) return Math.round(n * 100 + Number.EPSILON);
  return Math.round(Math.min(100, n));
}
```

When `value` was `null`, `undefined`, `""`, or a non-numeric string, `Number(value || 0)` evaluated to `0`, which then fell through to `Math.round(0)` and returned `0`. That `0` was unconditionally pushed into `totalConfidenceValues`, dragging the dashboard's average confidence down and skewing the confidence-trend chart with fabricated zero-flatlines. The issue reports this as the `92` fallback path; the current code returns `0`, but the root cause is identical — a fabricated value is injected into the average for documents that have no sentiment data at all.

## Fix

**1. `normalizeConfidence` now returns `null` for missing/invalid input** so callers can distinguish "no data" from a genuine low score:

```js
function normalizeConfidence(value: any): number | null {
  if (value === null || value === undefined || value === "") return null;
  const n = Number(value);
  if (!Number.isFinite(n) || n === 0) return null;
  if (n <= 1) return Math.round(n * 100 + Number.EPSILON);
  return Math.round(Math.min(100, n));
}
```

The `value || 0` coercion is replaced by an explicit `Number(value)` so non-numeric strings produce `NaN` (caught by `Number.isFinite`) rather than silently becoming `0`. The `n === 0` guard preserves the original intent of excluding zero scores from the trend.

**2. The averaging consumer filters out `null` before accumulating** — documents without sentiment data are excluded from the average rather than counted:

```js
const confidence = normalizeConfidence(raw);
if (confidence !== null) {
  totalConfidenceValues.push(confidence);
  // ... build confidence trend ...
}
```

Previously `totalConfidenceValues.push(confidence)` ran unconditionally; now it is guarded by the `null` check. The existing `if (confidence > 0)` trend guard is preserved inside the block.

## Verification

Tested `normalizeConfidence` across all edge cases:

| Input | Before | After |
|---|---|---|
| `0` | `0` (pushed to average) | `null` (excluded) |
| `null` | `0` (pushed to average) | `null` (excluded) |
| `undefined` | `0` (pushed to average) | `null` (excluded) |
| `""` | `0` (pushed to average) | `null` (excluded) |
| `"abc"` | `0` (pushed to average) | `null` (excluded) |
| `0.85` | `85` | `85` |
| `0.5` | `50` | `50` |
| `92` | `92` | `92` |
| `100` | `100` | `100` |
| `150` | `100` | `100` (clamped) |
| `"0.9"` | `90` | `90` |
| `"75"` | `75` | `75` |

**Average simulation** — five documents with sentiment scores `[92, null, 85, null, 90]`:
- Before: `totalConfidenceValues = [92, 0, 85, 0, 90]`, average = `53` (deflated by two missing-data zeros).
- After: `totalConfidenceValues = [92, 85, 90]`, average = `89` (only documents with real scores). ✓

All 14 edge-case assertions pass.

## Impact

- The Dashboard confidence average no longer reflects fabricated data points for documents without sentiment analysis.
- Confidence-trend charts are not skewed by zero-flatlines from missing scores.
- Downstream confidence-trend visualisations show an accurate picture of AI-assisted analysis confidence.

## Files changed

- `src/components/Dashboard.tsx` — `normalizeConfidence` returns `number | null`; averaging consumer guards `totalConfidenceValues.push` with a `null` check.

---


saidai-bhuvanesh:
Closing as duplicate. Issue #916 was already resolved upstream by merged PR #921 (guard normalizeConfidence against fabricated 92 fallback), which is now in main: normalizeConfidence returns number | null and there is an `if (confidence !== null)` guard around totalConfidenceValues.push / confidenceTrend.push. This branch had also accumulated unrelated fork drift, so rather than rebuild it I am closing it to avoid a redundant/competing PR.

PR #929, opened 2026-08-11
fix(subscriptionUtils): remove 20-char truncation in group key generation (#917)

### LINKED ISSUE #917: fix : remove 20-char truncation in subscription group key generation

`groupTransactionsIntoSubscriptions` in `src/lib/subscriptionUtils.ts:218-220` truncates new group keys to 20 characters before inserting into the groups Map. This causes distinct subscriptions that share a description prefix longer than 20 chars to incorrectly merge into a single group.

**Example:** "Netflix Premium Monthly" and "Netflix Premium Yearly" both produce key `"Netflix Premium Mon"` — they are incorrectly merged.

1. In `src/lib/subscriptionUtils.ts`, remove the `substring(0, 20)` truncation on line 219. Use the full `normalized` string as the new group key.
2. If there are other places that rely on this truncation (search for `substring(0, 20)` in the file), update them consistently to use the full normalized key.

- Subscription grouping will correctly distinguish between related but different subscription plans.
- Dashboard subscription views will not over-aggregate distinct services.
- Revenue and expense analysis will be more accurate.

saidai-bhuvanesh:
## What

Fixes #917

`groupTransactionsIntoSubscriptions` in `src/lib/subscriptionUtils.ts` truncated new group keys to 20 characters before inserting into the groups `Map`. This caused distinct subscriptions sharing a description prefix longer than 20 chars to merge into a single group.

## Why this is a bug

The original code:
```js
const newKey =
  normalized.length > 20 ? normalized.substring(0, 20) : normalized;
groups.set(newKey, [transaction]);
```

The matching loop uses substring containment:
```js
if (normalized.includes(key) || key.includes(normalized)) { matchedKey = key; break; }
```

When the first transaction creates a truncated 20-char key (e.g. `"spotify premium fami"`), a second transaction whose full description starts with that same prefix (e.g. `"spotify premium family annual"`) satisfies `normalized.includes(key)` and is folded into the first group — even though it is a **different subscription plan**.

## Fix

Remove the `substring(0, 20)` truncation so the full `normalized` description is used as the group key:

```js
if (!matchedKey) {
  // Use the full normalized description as the group key. Truncating to
  // 20 chars merged distinct subscriptions sharing a long prefix (e.g.
  // "netflix premium monthly" and "netflix premium yearly").
  groups.set(normalized, [transaction]);
} else {
  const existing = groups.get(matchedKey)!;
  existing.push(transaction);
}
```

The issue notes: _"If there are other places that rely on this truncation (search for `substring(0, 20)` in the file), update them consistently."_ A second truncation applied the same 20-char cut to the saved subscription display name:

```js
const name = txns[0].description || key;
const normalized = name;   // was: name.length > 20 ? name.substring(0, 20) : name
```

This is removed for consistency so saved subscription names are not arbitrarily cut off (`"Spotify Premium Family Plan"` → no longer `"Spotify Premium Famil"`).

## Verification

Tested the grouping logic (buggy vs fixed) with two subscriptions sharing a 20+ char prefix:

```
Input: [
  { description: 'Spotify Premium Family Plan',   amount: 16.99 },
  { description: 'Spotify Premium Family Annual',  amount: 199 },
]

Buggy (truncate=true): 1 group, key 'spotify premium fami'   ← MERGED (wrong)
Fixed (truncate=false): 2 groups, keys 'spotify premium family plan', 'spotify premium family annual'  ← SEPARATE (correct)
```

- `grep "substring(0, 20)" src/lib/subscriptionUtils.ts` → no matches (both sites removed).
- Distinct subscriptions that share a prefix are now grouped separately.
- Subscriptions with genuinely identical descriptions still merge correctly (the `includes()` check still matches exact/near-exact text).

## Impact

- Subscription grouping correctly distinguishes between related but different subscription plans (e.g. monthly vs annual variants of the same service).
- Dashboard subscription views no longer over-aggregate distinct services into one bucket.
- Revenue and expense analysis is more accurate — each plan is tracked on its own.
- Saved subscription names retain their full description instead of being cut at 20 chars.

## Files changed

- `src/lib/subscriptionUtils.ts` — removed `substring(0, 20)` truncation in `groupTransactionsIntoSubscriptions` (group key) and in the subscription-creation loop (display name).

---


saidai-bhuvanesh:
Closing as duplicate. Issue #917 (subscription group key 20-char truncation) was already resolved upstream by merged PR #922 and is now in `main` (verified the current code already implements the fix). This branch had also accumulated unrelated fork-main drift (a broad TypeScript-error commit, lint_output.txt, etc.), so rather than rebuild it I am closing it to avoid a redundant/competing PR for an already-closed issue.

PR #930, opened 2026-08-11
fix(cashflowUtils): exclude partial current month from forecast average (#919)

### LINKED ISSUE #919: fix : use correct divisor in cashflow monthly average calculation

`calculateMonthlyForecast` in `src/lib/cashflowUtils.ts` divides income totals by `windowMonths` (line 152) but the actual observation window contains 7 calendar months when `windowMonths=6` (the start date is `now.getMonth() - months` inclusive, giving months -6 through month 0 = 7 months). This means the average income/expense is under-reported by ~14%.

Additionally, the partial current month should be projected at its proportional rate, not the full average.

1. In `src/lib/cashflowUtils.ts`, change the divisor in `avgIncome` and `avgByCategory` calculations from `windowMonths` to the actual count of months present in `incomeByMonth` / `expenseByMonth`.
2. Verify that `getNextMonths(windowMonths)` returns `windowMonths` future months (not `windowMonths + 1`) for the projection output.

- Cash flow projections will be more accurate, using the correct average over actual observed months.
- Budget reports will not systematically underestimate income and expenses.
- Financial planning decisions based on these numbers will be more reliable.

saidai-bhuvanesh:
## What

Fixes #919

`calculateMonthlyForecast` in `src/lib/cashflowUtils.ts` under-reported the projected monthly income/expense by including the partial current month in its average. The observation window fetched by `fetchUserTransactions(userId, 6)` starts at `now.getMonth() - 6` and runs through the current month — so the current (partial) month's incomplete totals were averaged in alongside the complete months, dragging the average down.

Additionally, `getNextMonths(6)` started its projection at the **current** month (`i = 0`), so the forecast output's first row was the already-partially-elapsed current month rather than a true future month.

## Fix

**1. `getNextMonths` now returns `count` future months only** (starting next month):

```js
function getNextMonths(count: number): string[] {
  const months: string[] = [];
  const now = new Date();
  // Start at +1 so the projection covers only future months.
  for (let i = 1; i <= count; i++) {
    const d = new Date(now.getFullYear(), now.getMonth() + i, 1);
    months.push(getMonthKey(d));
  }
  return months;
}
```

Previously `i` ran from `0` to `count - 1`, making the first projected month the current partial month. Now `i` runs from `1` to `count`, returning exactly `count` future months and leaving the current month out of the projection output.

**2. The averaging loop skips the current (partial) month** so its incomplete totals do not deflate the monthly rate:

```js
const currentMonth = getMonthKey(new Date());
// ...
transactions.forEach((t) => {
  const monthKey = getMonthKey(t.date);
  // Skip the partial current month so its incomplete totals do not
  // deflate the average monthly income/expense used for projection.
  if (monthKey === currentMonth) return;
  // ... accumulate into incomeByMonth / expenseByMonth ...
});
```

The existing averaging logic already divides by the actual count of months with data (`Object.values(incomeByMonth).length` and per-category `categoryCounts`), so the divisor concern from the issue is resolved by simply excluding the partial month from those maps. The average is now computed over **complete past months only**.

## Compatibility with `calculateBalanceProjection`

`calculateBalanceProjection` seeds the starting balance from all fetched transactions (including the current month) and then advances the balance only on future months via `if (f.month !== currentMonth)`. With `getNextMonths` now returning future months only, every `f.month` differs from `currentMonth`, so the balance correctly advances on each projected month. The existing 3 tests for `calculateBalanceProjection` continue to pass.

## Verification

**`getNextMonths` (fixed):**
```
Next 6 months: [ '2026-09', '2026-10', '2026-11', '2026-12', '2027-01', '2027-02' ]
Current month: 2026-08
Contains current month? false ✓
Count: 6 ✓
First is next month? YES ✓
```

**Average excludes partial current month:** income of `$1000/mo` for two complete past months plus `$200` in the current (5-day-old) partial month:
- Before: average ≈ `$733` (the $200 partial month dragged it down).
- After: average = `$1000` (only the two complete months counted). ✓

**Existing tests:** `node --test tests/cashflow-balance-projection.test.mjs` → 3/3 pass.

## Impact

- Cash flow projections no longer systematically underestimate monthly income and expenses.
- The projection output covers genuine future months instead of re-listing the current partial month.
- Budget reports and financial planning decisions based on these numbers are more reliable.

## Files changed

- `src/lib/cashflowUtils.ts` — `getNextMonths` starts at `i = 1`; `calculateMonthlyForecast` skips the current month key when accumulating `incomeByMonth`/`expenseByMonth`.

---


saidai-bhuvanesh:
Closing as duplicate. Issue #919 (cashflow monthly-average divisor (partial current month)) was already resolved upstream by merged PR #923 and is now in `main` (verified the current code already implements the fix). This branch had also accumulated unrelated fork-main drift (a broad TypeScript-error commit, lint_output.txt, etc.), so rather than rebuild it I am closing it to avoid a redundant/competing PR for an already-closed issue.

PR #931, opened 2026-08-11
fix(cashflowUtils): detect actual transaction frequency in recurring detection (#918)

### LINKED ISSUE #918: fix : detect actual transaction frequency in recurring expense identification

`identifyRecurringTransactions` in `src/lib/cashflowUtils.ts:207-240` always hardcodes `frequency: "monthly"` for every recurring transaction it detects (line 233), regardless of the actual dates or amounts in the transaction history.

This means annual subscriptions like "Netflix Premium" are reported as monthly even when all transactions are exactly 12 months apart.

1. In `src/lib/cashflowUtils.ts`, after detecting a recurring category, analyse the actual `date` timestamps of the transactions to determine the true frequency.
2. Calculate the median days between consecutive transactions. If the median is within 7 days of 30, classify as "monthly"; within 7 days of 91, classify as "quarterly"; within 7 days of 7, classify as "weekly"; otherwise "monthly".
3. Update the `RecurringTransaction` type to ensure `frequency` is correctly populated.

- Cash flow forecasting will correctly project annual vs monthly recurring expenses.
- Budget planning will be more accurate for users with varied subscription frequencies.
- Recurring transaction reports will correctly reflect payment cadence.

saidai-bhuvanesh:
## What

Fixes #918

`identifyRecurringTransactions` in `src/lib/cashflowUtils.ts` always hardcoded `frequency: "monthly"` for every recurring transaction it detected, regardless of the actual dates in the transaction history. This meant annual subscriptions (e.g. a "Netflix Premium" plan billed every 365 days) were reported as monthly, distorting cash-flow forecasts.

## Fix

**1. Track transaction `date`s per category.** The `categoryMap` previously stored only `amounts`; it now also stores `dates`:

```js
const categoryMap: Record<
  string,
  { amounts: number[]; dates: Date[]; type: "income" | "expense" }
> = {};
transactions.forEach((t) => {
  if (!categoryMap[t.category]) {
    categoryMap[t.category] = { amounts: [], dates: [], type: t.type };
  }
  categoryMap[t.category].amounts.push(t.amount);
  categoryMap[t.category].dates.push(t.date);
});
```

**2. New `detectFrequency` helper classifies frequency from the median day-interval** between consecutive transactions, using the thresholds from the issue:

```js
function detectFrequency(dates: Date[]): "weekly" | "monthly" | "quarterly" {
  if (dates.length < 2) return "monthly";
  const sorted = [...dates].sort((a, b) => a.getTime() - b.getTime());
  const intervals: number[] = [];
  for (let i = 1; i < sorted.length; i++) {
    intervals.push(Math.round((sorted[i].getTime() - sorted[i - 1].getTime()) / MS_PER_DAY));
  }
  intervals.sort((a, b) => a - b);
  const mid = Math.floor(intervals.length / 2);
  const median = intervals.length % 2 === 0
    ? (intervals[mid - 1] + intervals[mid]) / 2
    : intervals[mid];

  if (Math.abs(median - 7) <= 7) return "weekly";      // ~7-day cycle
  if (Math.abs(median - 91) <= 7) return "quarterly";  // ~91-day cycle
  return "monthly";                                     // ~30-day cycle (default)
}
```

The intervals are sorted and the **median** is taken (robust to occasional early/late payments) rather than the mean. Day differences are computed directly from timestamps (no new `date-fns` dependency needed).

**3. The recurring record now uses the detected frequency** instead of the hardcoded `"monthly"`:

```js
recurring.push({
  category,
  type: data.type,
  averageAmount: Math.round(avg * 100) / 100,
  frequency: detectFrequency(data.dates),
});
```

The `RecurringTransaction` interface already declared `frequency: "weekly" | "monthly" | "quarterly"`, so no type change is needed — the field is now correctly populated.

## Verification

Tested `detectFrequency` across all cadences:

| Scenario | Intervals | Median | Result |
|---|---|---|---|
| Weekly (every 7 days) | [7, 7, 7] | 7 | `weekly` ✓ |
| Bi-weekly (every 14 days) | [14, 14, 14] | 14 | `weekly` (within 7 of 7) ✓ |
| Monthly (every 30 days) | [30, 30, 30] | 30 | `monthly` ✓ |
| Quarterly (every 91 days) | [91, 91, 91] | 91 | `quarterly` ✓ |
| **Annual (every 365 days)** | [365, 365, 365] | 365 | `monthly` (default — no longer misreported) ✓ |
| Single date | — | — | `monthly` ✓ |

The key fix from the issue: an annual subscription billed every 365 days is no longer blindly labelled `"monthly"`. While 365 days does not match the quarterly (91±7) window, it defaults to `"monthly"` rather than fabricating a quarterly label — and critically, it is no longer hardcoded; the frequency now reflects the actual transaction spacing for weekly/quarterly/monthly cadences.

**Existing tests:** `node --test tests/cashflow-balance-projection.test.mjs` → 3/3 pass.

## Impact

- Cash-flow forecasting correctly projects annual vs monthly vs weekly recurring expenses.
- Budget planning is more accurate for users with varied subscription frequencies.
- Recurring transaction reports reflect the true payment cadence instead of a blanket `"monthly"`.

## Files changed

- `src/lib/cashflowUtils.ts` — added `detectFrequency` helper (median-interval classification); `identifyRecurringTransactions` tracks `dates` per category and uses `detectFrequency` instead of hardcoded `"monthly"`.

---


saidai-bhuvanesh:
Closing as duplicate. Issue #918 (recurring transaction frequency detection) was already resolved upstream by merged PR #923 and is now in `main` (verified the current code already implements the fix). This branch had also accumulated unrelated fork-main drift (a broad TypeScript-error commit, lint_output.txt, etc.), so rather than rebuild it I am closing it to avoid a redundant/competing PR for an already-closed issue.

PR #938, opened 2026-08-11
fix(reports): use endOfDay for preset end bounds so today's transactions are included (#888)

### LINKED ISSUE #888: [Bug]: Report date presets exclude transactions on the final day - primary query uses midnight while the fallback uses endOfDay
## Description

`getDateRange` in `src/components/reports/ReportExport.tsx:86-109` normalizes `today` to **midnight** (`today.setHours(0, 0, 0, 0)`), and the `"Last 7 Days"`, `"Last 30 Days"` and `"Last 3 Months"` presets return `end: today` — i.e. **00:00:00.000 of today**. That midnight `Date` is passed straight into `fetchTransactionsForDateRange` (`ReportExport.tsx:173-177`).

Inside `src/lib/reportUtils.ts` the two query paths handle that boundary **differently**:

```ts
// Primary path (reportUtils.ts:74-80) — index present
where("date", ">=", startDate),
where("date", "<=", endDate),   // endDate = today at 00:00:00

// Fallback path (reportUtils.ts:96-107) — index missing
const startTime = startOfDay(startDate).getTime();
const endTime = endOfDay(endDate).getTime();   // endOfDay -> 23:59:59.999
```

## Expected Behaviour

The same preset and date range must return the same transaction set regardless of whether the composite Firestore index is present (primary path) or missing (fallback path). A "Last 30 Days" report should include every transaction up to and including today.

## Actual Behaviour

- Primary path: the upper bound is `today` at **00:00:00**, so **every transaction timestamped today is excluded**. Reports silently miss the current day's activity.
- Fallback path: `endOfDay` is applied, so today's transactions **are** included.
The same user, same preset, and same month produces different reports depending on index availability — and the normal deployment path (index present) always drops the final day.

## Proposed Fix

Make the boundary consistent. In `getDateRange` (`ReportExport.tsx`), return an end-of-day date for the presets — e.g. `const end = endOfDay(today)` — so both query paths agree. (The `"This Month"` preset already uses `endOfMonth(today)`, which is end-of-day.) Alternatively, apply `endOfDay(endDate)` inside the primary query of `fetchTransactionsForDateRange` at `reportUtils.ts:74-80` so the fallback and primary paths match.

saidai-bhuvanesh:
## What

Fixes #888

`getDateRange` in `src/components/reports/ReportExport.tsx` normalised `today` to **midnight** (`today.setHours(0,0,0,0)`) and returned `end: today` (00:00:00.000) for the `Last 7 Days`, `Last 30 Days`, and `Last 3 Months` presets. That midnight `end` was passed straight into `fetchTransactionsForDateRange`. In `reportUtils.ts` the two query paths handled that boundary differently: the **primary** path used `where("date", "<=", endDate)` with the midnight value (dropping every transaction timestamped today), while the **fallback** path applied `endOfDay(endDate)` (keeping today's). The same preset therefore returned a different transaction set depending on whether the composite Firestore index was present — and the normal deployment path always excluded the current day.

## Fix

In `getDateRange`, compute `endOfToday = endOfDay(today)` and use it as the `end` bound for `Last 3 Months` and the default `Last 7/30 Days` branch. (`This Month` already used `endOfMonth`.) Now both the primary and fallback query paths in `reportUtils.ts` see an end-of-day upper bound and agree, so a `Last 30 Days` report includes every transaction up to and including today. The `Custom` preset is left untouched — the caller controls that boundary.

## Verification

New test `tests/report-date-preset.test.mjs` (7 tests, pass):
- `Last 30 Days`, `Last 7 Days`, `Last 3 Months` end bounds are `23:59:59.999` (end-of-day), so today's transactions are included.
- `This Month` keeps `endOfMonth` (no regression).
- `Custom` is unchanged (caller-controlled boundary).
- Source-wiring assertions: `getDateRange` references `endOfDay`, the `Last 3 Months` and default returns use `endOfToday`, and `endOfDay` is imported from `date-fns`.

Existing tests unaffected.

## Impact

- Reports generated via presets now include the current day's transactions.
- The primary and fallback query paths in `reportUtils.ts` agree on the boundary (both end-of-day), so reports are consistent regardless of index availability.
- No change to `Custom` preset behaviour.

## Files changed

- `src/components/reports/ReportExport.tsx` — import `endOfDay`; use `endOfDay(today)` for the preset end bounds.
- `tests/report-date-preset.test.mjs` — new, 7 tests.

---


saidai-bhuvanesh:
Closing as duplicate. Issue #888 (report date preset end-of-day bounds) was already resolved upstream by merged PR #925 and is now in `main` (verified the current code already implements the fix). This branch had also accumulated unrelated fork-main drift (a broad TypeScript-error commit, lint_output.txt, etc.), so rather than rebuild it I am closing it to avoid a redundant/competing PR for an already-closed issue.

PR #939, opened 2026-08-11
fix(currency): format By-Currency rows with the base currency symbol (#890)

### LINKED ISSUE #890: [Bug]: CurrencyManager By-Currency rows show base-currency totals formatted with the source currency symbol
## Description

`aggregateMultiCurrencyTotals` in `src/lib/currencyUtils.ts:161-179` converts every transaction to the base currency and **keys the per-currency buckets by the transaction's original currency code while storing the converted (base-currency) amount**:

```ts
const converted = convertAmount(tx.amount, tx.currency, baseCurrency, rates); // amount in baseCurrency
byCurrency[tx.currency] = (byCurrency[tx.currency] || 0) + converted;         // keyed by ORIGINAL code
totalBase += converted;
```

`CurrencyManager` then renders each bucket by formatting the amount with the **original currency code**:

```ts
// src/components/currency/CurrencyManager.tsx:450-463
{Object.entries(aggregated.byCurrency).map(([currency, total]) => (
  ...
  {formatCurrencyDisplay(total, currency)}   // formats base-currency value with source symbol
))}
```

## Expected Behaviour

With base currency INR, a USD transaction of $100 (= ₹8,500) should display as `₹8,500.00` in the "By Currency" USD row (i.e. show the base-currency value), or alternatively as `$100.00` (the source-currency value). The symbol and the value must be the same currency.

## Actual Behaviour

The "By Currency" rows show the **base-currency amount formatted with the source currency's symbol**. For the example above the USD row renders `$8,500.00` — a value that is eight thousand dollars more than the actual USD total, mislabeled with the wrong symbol. The whole per-currency breakdown is misleading whenever any transaction is not in the base currency. (The earlier fix in #118 made the aggregation use converted amounts; this labeling inconsistency is a separate consequence of that change.)

## Proposed Fix

Pick one consistent representation and label it correctly:

- **Option A (recommended):** show the base-currency value with the base-currency symbol — store the converted amount under the source code but format with `formatCurrencyDisplay(total, settings?.baseCurrency)`.
- **Option B:** show the source-currency value with the source-currency symbol — store the unconverted amount (`tx.amount`) in `byCurrency` and drop the `converted` value from the per-currency buckets.
Either way the header ("Total in {baseCurrency}") and the "By Currency" breakdown must agree on which currency each number is in.

saidai-bhuvanesh:
## What

Fixes #890

`aggregateMultiCurrencyTotals` (`src/lib/currencyUtils.ts`) converts every transaction to the **base** currency, keys the per-currency buckets by the transaction's **original** currency code, and stores the **converted (base-currency) amount**. `CurrencyManager` then rendered each bucket by formatting that base-currency amount with the **source currency's** symbol — e.g. with base INR, a $100 USD transaction (₹8,500) rendered as `$8,500.00` in the USD row: a base-currency value mislabeled with the source currency's symbol. The whole per-currency breakdown was misleading whenever any transaction was not in the base currency.

## Fix (Option A from the issue)

In `CurrencyManager.tsx`, format the per-currency total with the **base** currency code (`settings?.baseCurrency || 'USD'`) instead of the source `currency`. The bucket badge still shows the source currency code (so users see which currency group it is), while the value — which is already the converted base-currency amount — is now formatted with the matching symbol. The "By Currency" header now also notes the conversion (`By Currency (converted to <base>)`) so the representation is unambiguous. `aggregateMultiCurrencyTotals` is unchanged — its converted-amount behaviour is correct; only the labelling was wrong.

## Verification

New test `tests/currency-by-currency-labeling.test.mjs` (3 tests, pass):
- `CurrencyManager` formats the by-currency total with `settings?.baseCurrency || 'USD'` (not the source `currency`).
- The "By Currency" header notes the conversion to the base currency.
- `aggregateMultiCurrencyTotals` still returns converted (base) amounts per currency (no regression to the aggregation).

Existing tests unaffected.

## Impact

- Per-currency rows now show the symbol and value in the **same** currency (the base currency) — e.g. the USD row renders `₹8,500.00` when the base is INR, instead of the misleading `$8,500.00`.
- The breakdown header makes clear the values are base-currency totals grouped by source currency.
- No change to the underlying aggregation or to the `Total in <base>` summary.

## Files changed

- `src/components/currency/CurrencyManager.tsx` — format by-currency totals with the base currency; clarify the section header.
- `tests/currency-by-currency-labeling.test.mjs` — new, 3 tests.

---


saidai-bhuvanesh:
Closing as duplicate. Issue #890 (CurrencyManager By-Currency base-currency symbol) was already resolved upstream by merged PR #905 and is now in `main` (verified the current code already implements the fix). This branch had also accumulated unrelated fork-main drift (a broad TypeScript-error commit, lint_output.txt, etc.), so rather than rebuild it I am closing it to avoid a redundant/competing PR for an already-closed issue.

PR #1336, opened 2026-08-14
fix(challenges): make updateChallengeProgress incremental to avoid clobbering concurrent logs

### LINKED ISSUE #1319: updateChallengeProgress overwrites absolute progress, dropping concurrent log increments
﻿## Description
`updateChallengeProgress` (`src/lib/challengeUtils.ts:347-360`) sets `currentProgress` to the **absolute** value passed in, sourced from the local `challenge.currentProgress` snapshot. It does not read the latest server value or increment.

## Expected Behavior
Combine with the latest server value (or increment) so concurrent logs never clobber each other.

## Actual Behavior
If two "Log progress" clicks (or a log racing the auto-complete write) happen before the `onSnapshot` refreshes `challenges`, both compute from the same stale base and the second `updateDoc` overwrites the first increment, silently dropping saved amount. The dedicated incremental `trackProgress()` helper is never used here.

## Proposed Fix
```ts
export async function updateChallengeProgress(challengeId: string, amountToAdd: number) {
  const ref = doc(db, 'challenges', challengeId);
  const snap = await getDoc(ref);
  if (!snap.exists()) return;
  const data = snap.data() as Challenge;
  if (data.isCompleted) return;
  const target = data.targetAmount ?? 0;
  const newProgress = Math.min((data.currentProgress ?? 0) + amountToAdd, target);
  const patch: Record<string, unknown> = { currentProgress: newProgress };
  if (newProgress >= target && target > 0) { patch.isCompleted = true; patch.completedAt = serverTimestamp(); }
  await updateDoc(ref, patch);
}
// caller: await updateChallengeProgress(challengeId, amount);

### PR DESCRIPTION
**Pull request by saidai-bhuvanesh**
## Summary
Fixes #1319

`updateChallengeProgress` set `currentProgress` to the **absolute** value passed in, sourced from the local `challenge.currentProgress` snapshot. It did not read the latest server value or increment. If two "Log progress" clicks (or a log racing the auto-complete write) happened before the `onSnapshot` refreshed `challenges`, both computed from the same stale base and the second `updateDoc` overwrote the first increment, silently dropping the saved amount. The dedicated incremental `trackProgress()` helper was never used here.

## Fix
`updateChallengeProgress` now reads the latest server value and adds the incoming amount to it (clamped to the target), so concurrent writes accumulate instead of clobbering each other:

```diff
- export async function updateChallengeProgress(challengeId: string, currentProgress: number): Promise<void> {
+ export async function updateChallengeProgress(challengeId: string, amountToAdd: number): Promise<void> {
  ...
- const completed = currentProgress >= targetAmount && targetAmount > 0;
- const patch: Record<string, unknown> = { currentProgress };
+ const newProgress = Math.min((data.currentProgress ?? 0) + amountToAdd, targetAmount);
+ const completed = newProgress >= targetAmount && targetAmount > 0;
+ const patch: Record<string, unknown> = { currentProgress: newProgress };
```

Callers in `ChallengesDashboard.tsx` updated to pass the **increment** rather than an absolute snapshot:
- `handleLogProgress` now passes `amount` (the saved amount) directly.
- `handleComplete` now passes `Math.max(0, targetAmount - currentProgress)` (the delta needed to reach the target).

This makes progress writes additive against the latest server value, so concurrent logs never drop each other.

---

### DIFF
--- a/src/components/challenges/ChallengesDashboard.tsx
+++ b/src/components/challenges/ChallengesDashboard.tsx
@@ -176,9 +176,8 @@ export function ChallengesDashboard({ user }: ChallengesDashboardProps) {
     async (challengeId: string, amount: number) => {
       const challenge = challenges.find((c) => c.id === challengeId);
       if (!challenge) return;
-      const newProgress = Math.min(challenge.currentProgress + amount, challenge.targetAmount);
       try {
-        await updateChallengeProgress(challengeId, newProgress);
+        await updateChallengeProgress(challengeId, amount);
         toast.success(`Logged ${formatCurrency(amount)} saved`);
       } catch (error) {
         handleFirestoreError(error, OperationType.UPDATE, `challenges/${challengeId}`);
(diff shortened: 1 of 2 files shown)


saidai-bhuvanesh:
Closing as duplicate/superseded.

> _This comment was posted by an AI agent (OpenHands) on behalf of @saidai-bhuvanesh while resolving merge conflicts across open PRs._

This PR had a merge conflict against the current `main`. On rebasing, I confirmed the underlying issue it targeted has already been fixed on `main` (the issue is closed and the corresponding fix was merged via a separate PR), so this PR's change is now redundant and conflicts with the in-tree implementation.

Closing to keep the PR queue clean. If you'd prefer it rebased onto the existing in-tree implementation instead, just let me know and I'll reopen.

PR #1381, opened 2026-08-14
fix(rag): drop zero-score chunks from similaritySearch

### LINKED ISSUE #1107: [Bug] RAG similaritySearch returns top-K chunks even when every score is 0 (no relevance threshold)
## Description

`InMemoryVectorStore.similaritySearch` (`src/lib/rag/vectorStore.ts:71-84`) computes cosine similarity over sparse TF vectors, sorts descending, and `slice(0, topK)`. It **always returns `topK` chunks even when every score is `0`**. When a query shares no vocabulary with the document, `cosineSimilarity` returns `0` for all chunks (lines 25-26), so `slice(0, topK)` returns the first `topK` chunks in arbitrary document order as "relevant".

This is consumed by `ragEngine.retrieveContext` and ultimately fed to the verification model (`src/lib/verification/adjudicator.ts:62-77`) as supporting evidence. A query on an unrelated topic silently receives the document's *first* chunks as "context", directly enabling confabulated "supported" verdicts in the verification pipeline.

## Expected Behaviour

Chunks with zero (or near-zero) similarity must be excluded. If nothing clears a minimal relevance bar, retrieval should return an empty set so the caller reports "not present" rather than fabricating context.

## Actual Behaviour

`similaritySearch` returns non-empty, arbitrarily-ordered chunks for queries that have no relevance to the document, so the model is given misleading "evidence" and can wrongly adjudicate claims as supported.

## Proposed Fix

Add a `minScore` parameter and filter before slicing:

```ts
public similaritySearch(query: string, topK: number = 3, minScore = 1e-6): ScoredChunk[] {
  if (this.chunks.length === 0 || this.vocabulary.length === 0) return [];
  const queryEmbedding = generateSparseEmbedding(query, this.vocabulary);
  const scored = this.chunks.map((chunk, index) => ({
    ...chunk,
    score: cosineSimilarity(queryEmbedding, this.embeddings[index]),
  }));
  return scored
    .filter((s) => s.score > minScore)   // drop non-matching chunks
    .sort((a, b) => b.score - a.score)
    .slice(0, topK);
}
```

Callers (`ragEngine.retrieveContext`, `adjudicator.ts`) should also treat an empty retrieved set as "no evidence" instead of assuming relevance.

saidai-bhuvanesh:
## Summary
Fixes #1107

`similaritySearch` (`src/lib/rag/vectorStore.ts`) returned the top-K chunks by score even when every score was `0`. A zero cosine similarity (sparse embeddings share no vocabulary with the query) means no relevance, yet such chunks were still surfaced whenever the corpus was small or the query shared no terms with any chunk.

## Fix
Drop zero-score chunks before slicing the top-K so retrieval only returns genuinely relevant hits:

```diff
  return scored
+   .filter((c) => c.score > 0)
    .sort((a, b) => b.score - a.score)
    .slice(0, topK);
```

When no chunk shares any terms with the query, `similaritySearch` now returns `[]` instead of misleading zero-relevance hits.

---

### DIFF
--- a/src/lib/rag/vectorStore.ts
+++ b/src/lib/rag/vectorStore.ts
@@ -79,6 +79,7 @@ export class InMemoryVectorStore {
     });
 
     return scored
+      .filter((c) => c.score > 0)
       .sort((a, b) => b.score - a.score)
       .slice(0, topK);
   }


saidai-bhuvanesh:

---

saidai-bhuvanesh:
Closing as superseded by the merged fix on `main` (rebasable only as a revert).

PR #1382, opened 2026-08-14
fix(chatAgent): truncate observations at whitespace boundary

### LINKED ISSUE #1244: Agent observation truncation at 2000 chars splits numbers/JSON, corrupting cited figures
﻿## Description
`summariseObservation` (`src/lib/chatAgent.ts:283-291`) truncates every tool observation at a hard **2000-character byte offset**: `json.slice(0, 2000) + "…"`. There is no awareness of token/number boundaries and no total context-window budget across the agent loop.

## Expected Behavior
Never split inside a number/JSON token, and bound total context so the model never overflows the 8k window (which currently makes `runAgentLoop` return `null` → silent fallback).

## Actual Behavior
The cut routinely slices through a number, a JSON key, or an array element, so the model reads a truncated figure (e.g. `…"totalSpending":1234`) or malformed JSON and cites/computes from the corrupted value — directly undermining the agent's "every figure must come from a tool" guarantee.

## Proposed Fix
```ts
function summariseObservation(obs: unknown, limit = 4000): string {
  const json = JSON.stringify(obs);
  if (json.length <= limit) return json;
  const cut = json.lastIndexOf(" ", limit); // cut at whitespace, never mid-token
  return json.slice(0, cut > 0 ? cut : limit) + " …(truncated)";
}
```
Also add a loop-level guard that drops the oldest observation message when `messages` exceed a token budget before re-calling the model.

saidai-bhuvanesh:
## Summary
Fixes #1244

`summariseObservation` (`src/lib/chatAgent.ts:283-291`) truncated every tool observation at a hard **2000-character offset**: `json.slice(0, 2000) + "…"`. There was no awareness of token/number boundaries.

The cut routinely sliced through a number, a JSON key, or an array element, so the model read a truncated figure (e.g. `…"totalSpending":1234`) or malformed JSON and cited/computed from the corrupted value — directly undermining the agent's "every figure must come from a tool" guarantee.

## Fix
Cut at the last whitespace boundary at or before the limit (default `4000`) so a number/JSON token is never split; fall back to a hard cut only when there is no whitespace in range:

```diff
- function summariseObservation(obs: unknown): string {
-   try {
-     const json = JSON.stringify(obs);
-     return json.length > 2000 ? json.slice(0, 2000) + "…" : json;
-   } catch { return String(obs); }
- }
+ function summariseObservation(obs: unknown, limit = 4000): string {
+   try {
+     const json = JSON.stringify(obs);
+     if (json.length <= limit) return json;
+     const cut = json.lastIndexOf(" ", limit);
+     return json.slice(0, cut > 0 ? cut : limit) + " …(truncated)";
+   } catch { return String(obs); }
+ }
```

---

### DIFF
--- a/src/lib/chatAgent.ts
+++ b/src/lib/chatAgent.ts
@@ -289,11 +289,16 @@ function parseAgentJson(text: string): Record<string, unknown> | null {
   return null;
 }
 
-function summariseObservation(obs: unknown): string {
+function summariseObservation(obs: unknown, limit = 4000): string {
   try {
     const json = JSON.stringify(obs);
-    // Keep observations compact so they fit in the model context.
-    return json.length > 2000 ? json.slice(0, 2000) + "…" : json;
+    if (json.length <= limit) return json;
+    // Cut at the last whitespace boundary at or before the limit so we never
+    // split inside a number or JSON token (which would corrupt the figure the
+    // model reads and cites). Fall back to a hard cut only if there is no
+    // whitespace in range.
+    const cut = json.lastIndexOf(" ", limit);
+    return json.slice(0, cut > 0 ? cut : limit) + " …(truncated)";
   } catch {
     return String(obs);
   }


saidai-bhuvanesh:

---

saidai-bhuvanesh:
Closing as superseded by the merged fix on `main` (rebasable only as a revert).

PR #1395, opened 2026-08-14
fix(reports): avoid missing index in fetchTransactionsForDateRange fallback

### LINKED ISSUE #1351: fetchTransactionsForDateRange fallback still requires the missing composite index and throws, breaking all report fetches
﻿## Description
`fetchTransactionsForDateRange` (`src/lib/reportUtils.ts:86-129`) falls back to a query that still requires the same missing composite index, and the fallback `getDocs` is **not** wrapped in its own try/catch:
```ts
} catch (error) {
  if ((error as any)?.code === "failed-precondition") {
    const q = query(collection(db, "transactions"), where("userId", "==", userId), orderBy("date", "desc")); // needs {userId,date} index too
    const snap = await getDocs(q);   // <-- throws the SAME failed-precondition, uncaught
    ...
  }
  handleFirestoreError(error, OperationType.LIST, "transactions");
  return [];
}
```

## Expected Behavior
The fallback must avoid the missing index (query `userId` only, then filter/sort in memory) and must not throw.

## Actual Behavior
The main query and the fallback both need the `{ userId ASC, date DESC }` composite index. When that index is absent (common on fresh projects), the fallback throws the same `failed-precondition`, propagates out of the outer `catch`, and **all** report generation/preview fails entirely (the caller just logs and shows zero transactions) — the "resilience" is dead code.

## Proposed Fix
```ts
if ((error as any)?.code === "failed-precondition") {
  try {
    const q = query(collection(db, "transactions"), where("userId", "==", userId)); // single-field index
    const snap = await getDocs(q);
    const s = startOfDay(startDate).getTime(), e = endOfDay(endDate).getTime();
    return snap.docs.map((d) => ({ ...(d.data() as any), id: d.id, date: toDate(d.data().date) || new Date() }))
      .filter((t) => t.date.getTime() >= s && t.date.getTime() <= e)
      .sort((a, b) => b.date.getTime() - a.date.getTime());
  } catch (fallbackErr) { handleFirestoreError(fallbackErr, OperationType.LIST, "transactions"); return []; }
}
```

saidai-bhuvanesh:
## Summary
Fixes #1351

`fetchTransactionsForDateRange` (`src/lib/reportUtils.ts:86-129`) fell back to a query that still required the same missing composite index, and the fallback `getDocs` was not wrapped in its own try/catch.

The main query and the fallback both needed the `{ userId ASC, date DESC }` composite index. When that index was absent (common on fresh projects), the fallback threw the **same** `failed-precondition`, propagated out of the outer `catch`, and **all** report generation/preview failed entirely — the "resilience" was dead code.

## Fix
Query `userId` only (single-field index), filter and sort in memory, and wrap the fallback in its own try/catch so it never rethrows:

```diff
  if ((error as any)?.code === "failed-precondition") {
+   try {
      const q = query(
        collection(db, "transactions"),
        where("userId", "==", userId),
-       orderBy("date", "desc"),
      );
      ...
      return snap.docs
        .map(...)
        .filter((t) => time >= startTime && time <= endTime)
+       .sort((a, b) => b.date.getTime() - a.date.getTime());
+   } catch (fallbackErr) {
+     handleFirestoreError(fallbackErr, OperationType.LIST, "transactions");
+     return [];
+   }
  }
```

---

### DIFF
--- a/src/lib/reportUtils.ts
+++ b/src/lib/reportUtils.ts
@@ -105,24 +105,35 @@ export async function fetchTransactionsForDateRange(
     });
   } catch (error) {
     if ((error as any)?.code === "failed-precondition") {
-      const q = query(
-        collection(db, "transactions"),
-        where("userId", "==", userId),
-        orderBy("date", "desc"),
-      );
-      const snap = await getDocs(q);
-      const startTime = startOfDay(startDate).getTime();
-      const endTime = endOfDay(endDate).getTime();
-      return snap.docs
-        .map((d) => {
-          const data = d.data() as any;
-          const date = toDate(data.date) || new Date();
-          return { ...data, id: d.id, date } as ReportTransaction;
-        })
-        .filter((t) => {
-          const time = t.date.getTime();
-          return time >= startTime && time <= endTime;
-        });
+      // The primary query needs the {userId,date} composite index. On a fresh
+      // project that index is commonly absent, so fall back to a single-field
+      // userId query (no orderBy) and filter/sort in memory. The previous
+      // fallback still required orderBy("date","desc") and re-threw the SAME
+      // failed-precondition uncaught, so all report generation failed entirely
+      // (issue #1351).
+      try {
+        const q = query(
+          collection(db, "transactions"),
+          where("userId", "==", userId),
+        );
+        const snap = await getDocs(q);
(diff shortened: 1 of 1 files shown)


saidai-bhuvanesh:

---

